PR #3, opened 2021-10-05
make also include build_deps stage

OfirMos:


Makefile
```diff
@@ -5,6 +5,7 @@ build_deps:
 	make -C deps/
 	
 libmr:
```
MeirShpilraien, line 7:
```suggestion
libmr: build_deps
```

MeirShpilraien (commented):
Small suggestion

PR #4, opened 2021-10-06
[WIP] Build system

rafie:


MeirShpilraien:
@rafie I do not think we need a complex build system here, its a small repository that does not even need to be deployed anywhere...

rafie:
@MeirShpilraien All OK, just a service to the community. :-) 
Closing.

PR #25, opened 2022-05-29
redis version bump

chayim:


gkorland (approved):

.circleci/config.yml
```diff
@@ -86,7 +86,7 @@ workflows:
           <<: *on-any-branch
           matrix:
             parameters:
-              redis_version: ["6.2.6", "7.0-rc2"]
+              redis_version: ["6.2.6", "7.0"]
```
MeirShpilraien, line 89: Shouldn't it  be 7.0.0?

MeirShpilraien (commented):
Small comment

.circleci/config.yml
chayim: both work - probably a result of the branch.  Committed the update.

MeirShpilraien (approved):

PR #46, opened 2023-09-07
Fix tls on Redis 7.2 oss cluster

MeirShpilraien:
On Redis 7.2 (specifically on 7.2-rc3) there was a [change](https://github.com/redis/redis/pull/12233) made that allow running on cluster with both, tls and none tls port.
As part of this change, `cluster slot` command will return the port base of the connection type (tls or not tls). A side effect of this change is that when `cluster slot` called using `RM_Call` it is always gets the none tls port (which break our support for tls).

To fix this, use `RedisModule_GetClusterNodeInfo` Redis module API to get the correct port.

Notice that this is a minimal fix, we should extend the Redis module API to provide us with all the information we need (including slot range) so we will not have to use `cluster slot` command at all.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -771,7 +771,7 @@ static void MR_RefreshClusterData(){
         // We need to get the port using the `RedisModule_GetClusterNodeInfo` API because on 7.2
         // invoking `cluster slot` from RM_Call will always return the none tls port.
         // For for information refer to: https://github.com/redis/redis/pull/12233
-        long long port = 0;
+        int port = 0;
         RedisModule_ThreadSafeContextLock(mr_staticCtx);
         RedisModule_GetClusterNodeInfo(mr_staticCtx, nodeId, NULL, NULL, &port, NULL);
         RedisModule_ThreadSafeContextUnlock(mr_staticCtx);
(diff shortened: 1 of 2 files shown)


src/cluster.c
```diff
@@ -770,6 +768,15 @@ static void MR_RefreshClusterData(){
         memcpy(nodeIp, ip, ipLen);
         nodeIp[ipLen] = '\0';
 
+        // We need to get the port using the `RedisModule_GetClusterNodeInfo` API because on 7.2
+        // invoking `cluster slot` from RM_Call will always return the none tls port.
+        // For for information refer to: https://github.com/redis/redis/pull/12233
+        long long port = 0;
+        RedisModule_ThreadSafeContextLock(mr_staticCtx);
```
GuyAv46, line 775: Do we want to lock and unlock inside the loop? Does it matter?

GuyAv46 (approved):

src/cluster.c
MeirShpilraien: Its a loop of few elements that should almost not happened, only on topology changes, so I do not think it is that important.

GuyAv46 (approved):

PR #53, opened 2023-12-13
Fix compile issues

tezc:
- Missing include guard for common.h fails the build on my local
- [redisAsyncSetConnectCallback()](https://github.com/RedisGears/LibMR/blame/d87f186bf62bbb2fb946853bbd7052be4c359aa1/src/cluster.c#L595) expects const pointer argument in the callback. Currently, compiler prints a warning. Added a cast to workaround the issue.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -513,7 +513,7 @@ error:
     return NULL;
 }
 
-static void MR_OnConnectCallback(struct redisAsyncContext* c, int status){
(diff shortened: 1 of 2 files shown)


tezc:
@MeirShpilraien Could you take a look sometime? Tried to pull latest libmr into timeseries and got two compile issues. Thank you.

YaacovHazan (approved):

MeirShpilraien (approved):

PR #54, opened 2024-11-22
Mod 8123 add acl support

iddm:


Cargo.toml
```diff
@@ -16,15 +15,17 @@ libc = "0.2"
...
+bindgen = "0.70"
 
 [lib]
 crate-type = ["rlib"]
 name = "mr"
 path = "rust_api/lib.rs"
 
 [features]
-default = ["min-redis-compatibility-version-6-0"]
+# default = ["min-redis-compatibility-version-6-0"]
```
iddm, line 26: Can I remove it, or should we leave it?

src/redismodule-rlec.h
```diff
@@ -0,0 +1,96 @@
+#pragma once
```
iddm, line 1: @MeirShpilraien I know you don't like this film, and I agree we shouldn't invent those and maintain them. But this is how it has been in the RedisModulesSDK repository, which we use in some other projects as well. The problem here is that some symbols are only available in the RLEC version and some in the opensource. I think, this is why we have two files. But I am not knowledgeable enough when it comes to RLEC to know what and how exactly to do here to avoid this and do better. You may suggest any other way.

MeirShpilraien: Can you specify which API you need from this file?

MeirShpilraien:
I also do not see that you are actually using the new username on auth command here: https://github.com/RedisGears/LibMR/blob/master/src/cluster.c#L566
Am I missing something? I also do not see tests for it?

iddm:
> I also do not see that you are actually using the new username on auth command here: https://github.com/RedisGears/LibMR/blob/master/src/cluster.c#L566 Am I missing something? I also do not see tests for it?

Yes, I am still working on adding the tests for this. I wanted to make the basic code review first. When porting the changes from my old PR with the formatting, it seems I have forgotten to port this.

eranhd (commented):
@iddm please clean ups unrelated ACL changes

tests/mr_test_module/pytests/test_basic.py
```diff
@@ -78,3 +78,29 @@ def testRemoteTaskOnAllShards(env, conn):
     for i in range(100):
         conn.execute_command('del', 'doc%d' % i)
     env.expect('lmrtest.dbsize').equal(0)
+
+@MRTestDecorator()
+def testAclSetting(env, conn):
```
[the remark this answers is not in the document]
iddm: I had already tested it manually, and it would require changing the infra code a bit. Let's see how I can do that.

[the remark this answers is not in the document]
iddm: It is difficult to test it this way as the testing infrastructure requires changes, which I need to investigate. We cannot ask the RLTest here to spawn a redis-server with the given arguments; we do that externally from the outside, from the `run_tests.sh` script. This is why I had tested it manually before you asked, last week.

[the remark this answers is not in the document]
eranhd: Ok, but we can't leave it without a component test, right?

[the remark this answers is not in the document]
iddm: Added the test.

iddm:
[The phase two PR](https://github.com/RedisTimeSeries/RedisTimeSeries/pull/1657) is now updated to use this version of LibMR but as a branch, of course.

.github/workflows/macos.yml
```diff
@@ -8,35 +8,36 @@ on:
...
-        cd tests/mr_test_module/pytests/
-        DEBUG=1 ./run_full_tests.sh
+      run: brew install automake libtool
+    - name: Build the dependencies
+      run: make build_deps
+    - name: Default tests
+      run: make run_tests
+    - name: SSL tests
+      run: make run_tests_ssl
+    # - name: Valgrind tests
```
MeirShpilraien, line 42: We need to return valgrind run

src/cluster.c
```diff
@@ -36,6 +36,58 @@
...
+                                        ? LIBMR_ACL_COMMAND_CATEGORY_NAME           \
+                                        : acls " " LIBMR_ACL_COMMAND_CATEGORY_NAME; \
+            if (RedisModule_SetCommandACLCategories(command, categories) ==         \
+                REDISMODULE_OK) {                                                   \
+                result = true;                                                      \
+            }                                                                       \
+        }                                                                           \
+    }                                                                               \
+    result;                                                                         \
+  })
```
MeirShpilraien, line 64: Any reason why not to put it in a function?

src/cluster.c
```diff
@@ -562,9 +616,13 @@ static void MR_OnConnectCallback(const struct redisAsyncContext* c, int status){
         }
 
         RedisModule_Log(mr_staticCtx, "notice", "connected : %s:%d, status = %d\r\n", c->c.tcp.host, c->c.tcp.port, status);
-        if(n->password){
+
+        if (n->username && n->password) {
+            redisAsyncCommand((redisAsyncContext*)c, NULL, NULL, "AUTH %s %s", n->username, n->password);
```
MeirShpilraien, line 621: Lets not forget to create an issue to handle failure here

build.rs
```diff
@@ -21,29 +21,23 @@ fn main() {
...
+    let mut command = Command::new("make");
+
+    command.env(
+        "MODULE_NAME",
+        std::env::var("MODULE_NAME").expect("module name was not given"),
+    );
+
+    if let Ok(acl_category_name) = std::env::var("LIBMR_ACL_COMMAND_CATEGORY_NAME") {
+        command.env("LIBMR_ACL_COMMAND_CATEGORY_NAME", acl_category_name);
+    }
```
MeirShpilraien, line 33: I do not see us using it from the env var anywhere, so why is it needed? Am I missing something?

src/redismodule-rlec.h
MeirShpilraien: Can we try to remember and see if we can avoid this extra file?

tests/mr_test_module/Cargo.toml
```diff
@@ -1,15 +1,13 @@
...
 authors = ["meir@redislabs.com <meir@redislabs.com>"]
-edition = "2018"
+edition = "2021"
 license = "LicenseRef-RSALv2 OR SSPL-1.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-#redis-module = { version="0.22.0", features = ["experimental-api"]}
-redis-module = { git = "https://github.com/RedisLabsModules/redismodule-rs", branch = "master" }
```
MeirShpilraien, line 12: How come we do not need it?

tests/mr_test_module/pytests/common.py
```diff
@@ -95,18 +96,43 @@ def verifyClusterInitialized(env):
...
             test_name = '%s:%s' % (inspect.getfile(test_function), test_function.__name__)
             if skipTest and not runSkipTests():
                 raise unittest.SkipTest()
+            envArgs['moduleArgs'] = moduleArgs or None
             env = Env(**envArgs)
+            if skipOnVersionLowerThan:
+                env.skipOnVersionSmaller(skipOnVersionLowerThan)
+            if redisConfigFileContent:
+                envArgs['redisConfigFile'] = create_config_file(redisConfigFileContent) if redisConfigFileContent else None
+                env = Env(**envArgs)
```
MeirShpilraien, line 130: You are creating 2 env's here

tests/mr_test_module/pytests/test_basic.py
```diff
@@ -78,3 +78,48 @@ def testRemoteTaskOnAllShards(env, conn):
...
+    Tests that LibMR doesn't work when the user provided for it doesn't
+    have the necessary permissions to run the LibMR commands.
+    '''
+
+    # This should fail as the LibMR will attempt to connect to the
+    # shards using the "baduser" user, which doesn't have the necessary
+    # permissions to run the LibMR commands.
+    if env.isCluster():
+        with pytest.raises(Exception):
+            initialiseCluster(env)
```
MeirShpilraien, line 125: Can we verify some content of the exception maybe to make sure we failed where we expect to fail?

tests/mr_test_module/src/lib.rs
```diff
@@ -847,11 +849,29 @@ impl Drop for KeysReader {
...
+        2 => {
+            let username = args[0].to_string();
+            let password = args[1].to_string();
+            mr_init(ctx, 5, Some(username.as_str()), Some(password.as_str()));
+        }
+        _ => {
+            redis_module::logging::log_warning("Invalid number of arguments. Expected 0, 1, or 2.");
+            return Status::Err;
+        }
+    }
```
MeirShpilraien, line 874: Can we change it to use iterator on the args slice?

tests/mr_test_module/pytests/test_basic.py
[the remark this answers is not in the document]
MeirShpilraien: Why not change this test as well to use a dedicated libmr user and deny libmr commands on the default user?

tests/mr_test_module/pytests/common.py
```diff
@@ -117,15 +143,11 @@ def test_func():
...
-            env.broadcast('MRTESTS.REFRESHCLUSTER')
-            if env.isCluster():
-                # make sure cluster will not turn to failed state and we will not be
-                # able to execute commands on shards, on slow envs, run with valgrind,
-                # or mac, it is needed.
-                env.broadcast('CONFIG', 'set', 'cluster-node-timeout', '120000')
-                env.broadcast('MRTESTS.FORCESHARDSCONNECTION')
-                with TimeLimit(2):
-                    verifyClusterInitialized(env)
+            if commandsBeforeClusterStart:
```
MeirShpilraien, line 146: Where do you use this new option?

MeirShpilraien (commented):
:+1: 
Added some comments

.github/workflows/macos.yml
iddm: I understand that, but this is not a top-priority, I'd say. If you don't see any resource leakage in the changes I made, can't we consider it okay?

tests/mr_test_module/Cargo.toml
iddm: Cuz we simply use what the libmr-rust crate reexports. A common practice.

tests/mr_test_module/pytests/common.py
iddm: Yes, cuz it is impossible to know in advance the redis version the `Env` object will try to use.

.github/workflows/macos.yml
MeirShpilraien: No, we can not disable valgrind run, if you think it failed even before your changes then this is probably top priority to fix.

tests/mr_test_module/src/lib.rs
iddm: I don't think so; I tried to achieve that a year ago in another project, but it was a mess. The problem is that we need to know the number of strings we receive, and only then we know what is what. Are you sure we need to think too much about the test code anyway?

tests/mr_test_module/pytests/common.py
iddm: I used to use it before I started using the Redis config file content. If you want, I can delete it now as I don't use it anymore.

.github/workflows/macos.yml
iddm: I think it didn't fail, that's what I meant.

tests/mr_test_module/src/lib.rs
MeirShpilraien: Why can't it be something like:
```rust
let args_iter = args.iter();
let user = args_iter.next().map(|v| v.to_string());
let pass = args_iter.next().map_or_else(|| Some("password"), |v| v.to_string());
mr_init(ctx, 5, user, pass);
```

tests/mr_test_module/pytests/test_basic.py
iddm: I'll check once more, but we probably cannot, as there are no errors thrown neither by LibMR nor the redis server, there is just a timeout due to the shards not reaching the "connected" state.

build.rs
iddm: You don't see it because it is optional.

```c
#ifndef LIBMR_ACL_COMMAND_CATEGORY_NAME
#define LIBMR_ACL_COMMAND_CATEGORY_NAME                                        \
  "_" xstr(MODULE_NAME) "_libmr_internal"
#endif
```

.github/workflows/macos.yml
MeirShpilraien: So why you disabled it?

build.rs
MeirShpilraien: But this env var is not eventually given as `-DLIBMR_ACL_COMMAND_CATEGORY_NAME` argument to the compiler.
So it is basically ignored. Or am I missing something?

src/cluster.c
iddm: To avoid the manual heap allocation and concatenation of the strings `acls` and `LIBMR_ACL_COMMAND_CATEGORY_NAME`, this can be done at compile time as both are literals by:
```c
acls " " LIBMR_ACL_COMMAND_CATEGORY_NAME
```

There also isn't much code, in my personal opinion, to be bothered about the inlining; it is never called from within a loop and is quite simple to be optimised, I guess, but I haven't seen the actual optimisation outcome, I am just guessing it is done sufficiently. Also, again, this macro is used just a bunch of times when we register the commands, so it shouldn't cause any runtime concerns. The only concern here, in my opinion, is the readability. But again, it is too short to cause a panic attack, in my opinion.. :-)

tests/mr_test_module/pytests/common.py
MeirShpilraien: So lets move this check bellow, it is wrong to create 2 envs at the same test.

tests/mr_test_module/pytests/test_basic.py
MeirShpilraien: OK, so lets verify that the exception is a timeout, maybe add some message state that shards did not reach connect state.

build.rs
iddm: I just said it was optional :-) The real modules which use LibMR may want to use it whenever they want. If they don't use, the module name will be used as a prefix, just as it is done with RediSearch.

.github/workflows/macos.yml
iddm: I hope you can see why now (the CI failed).

```
 DEBUG=1 /home/runner/work/LibMR/LibMR/tests/mr_test_module/pytests/run_tests.sh -t ./pytests/ -V --vg-suppressions ./leakcheck.supp
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/RLTest/redis_std.py", line 347, in getInformationAfterDispose
    with open(os.path.join(self.dbDirPath, self._getFileName(role, '.log'))) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: './logs/395029cc5bbf42e8aec420585830f653.master-1-test_errors_testInternalCommandFromScript-oss.log'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/RLTest/__main__.py", line 1026, in <module>
    main()
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/RLTest/__main__.py", line 1022, in main
    RLTest().execute()
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/RLTest/__main__.py", line 959, in execute
    run_jobs_main_thread(jobs)
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/RLTest/__main__.py", line 916, in run_jobs_main_thread
    done += self.run_single_test(test, on_timeout)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/RLTest/__main__.py", line 843, in run_single_test
    verboseInfo['after_dispose'] = lastEnv.getInformationAfterDispose()
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/RLTest/env.py", line 286, in getInformationAfterDispose
    return self.envRunner.getInformationAfterDispose()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/RLTest/redis_std.py", line 349, in getInformationAfterDispose
    except os.FileNoteFoundError:
           ^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'os' has no attribute 'FileNoteFoundError'
Error 111 connecting to localhost:6379. Connection refused.
	❌  (FAIL):	'Error 111 connecting to localhost:6379. Connection refused.' should contain 'command is not allowed from script'	test_errors.py:27
	❌  (FAIL):	'Error 111 connecting to localhost:6379. Connection refused.' should contain 'command is not allowed from script'	test_errors.py:28
	❌  (FAIL):	'Error 111 connecting to localhost:6379. Connection refused.' should contain 'command is not allowed from script'	test_errors.py:29
	❌  (FAIL):	'Error 111 connecting to localhost:6379. Connection refused.' should contain 'command is not allowed from script'	test_errors.py:30
	❌  (FAIL):	'Error 111 connecting to localhost:6379. Connection refused.' should contain 'command is not allowed from script'	test_errors.py:31
	❌  (FAIL):	'Error 111 connecting to localhost:6379. Connection refused.' should contain 'command is not allowed from script'	test_errors.py:32
	❌  (FAIL):	'Error 111 connecting to localhost:6379. Connection refused.' should contain 'command is not allowed from script'	test_errors.py:33
	❌  (FAIL):	'Error 111 connecting to localhost:6379. Connection refused.' should contain 'command is not allowed from script'	test_errors.py:34
test_errors:testInternalCommandFromScript:
	[FAIL]
	```
I don't have time to bother to fix it now, we were promised to deliver a long time ago. If this is a major concern in your opinion and you think the new changes I made introduce any leaks or something, sure, let's spend time on it. But if it has been here for a long time without any issues, I think we can't address it later. I remember having a similar issue in the other projects, it was related to RLTest or redis-py or something, can't really remember. Ozan might.

src/cluster.c
iddm: https://github.com/RedisGears/LibMR/issues/55

build.rs
iddm: > But this env var is not eventually given as `-DLIBMR_ACL_COMMAND_CATEGORY_NAME` argument to the compiler. So it is basically ignored. Or am I missing something?

I see, you meant the use of it in the `build.rs` in particular. Yes, initially I was using it, but then removed. I'll remove the use of it now here as well as there is always an optional value. I apologise for the initial misunderstanding.

tests/mr_test_module/pytests/common.py
iddm: It is impossible to do it below.

tests/mr_test_module/pytests/test_basic.py
iddm: Yes, sure, just as I had already done in RedisTimeSeries.

tests/mr_test_module/src/lib.rs
iddm: Indeed, this way, it probably can. I had a little different use case, so I immediately rejected the possibility of trying to do this here for the lack of time. Thanks for the help.

tests/mr_test_module/pytests/common.py
iddm: I had another idea to avoid using the `Env` at all here, which was just to simply invoke the `Defaults.redis_binary` with the `--version` flag and parse the version before creating the `Env` object. This would be the solution I'd commit.

iddm: We don't have time for the best solutions here, again. So I did what I had to do, the company priorities are now to deliver, not to do it good.

tests/mr_test_module/pytests/test_basic.py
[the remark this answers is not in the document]
iddm: Because the purpose of the particular test you mentioned was to make sure the ACLs work in principle. And there is another test that confirms it doesn't work when it shouldn't either.

src/redismodule-rlec.h
iddm: I'd rather just add a RedisModulesSDKrepository as a submodule and rely on it there instead of remembering something. Let's maintain one code base and think of just one header. What it does and what doesn't - shouldn't be our problem, but the RedisModulesSDK problem.
NEW

So I'd ideally remove the redismodule.h and the other header from the libmr repository in favour of using the RedisModulesSDK directly.

tests/mr_test_module/pytests/common.py
MeirShpilraien: I see you changed it eventually.

tests/mr_test_module/pytests/test_basic.py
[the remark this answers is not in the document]
MeirShpilraien: If you keep use the default user it does not prove that libmr uses the user you gave, because the default is to use the default user.

src/redismodule-rlec.h
MeirShpilraien: OK, giving up, leave it as is.

.github/workflows/macos.yml
MeirShpilraien: We can not ignore this failure, even if it is not related to your changes we can not release without passing valgrind check.

tests/mr_test_module/pytests/test_basic.py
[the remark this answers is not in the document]
iddm: But the next test below this one uses a different user and confirms it fails with it. Doesn't this cover what you mean?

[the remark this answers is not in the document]
MeirShpilraien: I think that it is best to verify the exact case as the user would have use it. Our recommendation is going to be to set a dedicated user that has permissions to run the libmr category and deny access for this category for all other users. So lets test it this way.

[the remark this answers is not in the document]
iddm: Added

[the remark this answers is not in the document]
MeirShpilraien: What about removing this category for the default user?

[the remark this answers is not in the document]
iddm: I did this two days ago in RTS... Anyway, doing it here as well.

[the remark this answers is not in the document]
iddm: Added

MeirShpilraien (approved):

PR #56, opened 2025-01-22
Support internal commands

MeirShpilraien:
The PR makes changes to use the internal command capabilities introduced on: https://github.com/redis/redis/pull/13740.
The main idea is to use the new redismodule API that provides internal secret for internal shard communication.

This means that the following is no longer needed:
1. The new user configuration.
2. The internal ACL category

All the commands are marked with the `internal` flag (if we see that the internal commands feature is supported on the current Redis version we run on). In addition, the connection between the shards uses `INTERNALAUTH` with the given internal secret so it can run all the internal commands. Regular user can not run any internal command.

The PR require the following PR's to be merge:
1. https://github.com/redis/redis/pull/13740
2. https://github.com/RedisLabsModules/redismodule-rs/pull/407

The tests was modified to avoid giving password to the test module in case it detects that the internal commands feature exists. In such case, avoiding giving the password to the module, make sure that LibMR actually using the `INTERNALAUTH` command with the internal secret to authenticate.

todo:
- [x] Chnage redismodule-rs version to `master` once PR will be merged: https://github.com/RedisLabsModules/redismodule-rs/pull/407

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        redis_version: ["7.2", "7.4"]
+        redis_version: ["7.2", "7.4", "unstable"]
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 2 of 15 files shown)


MeirShpilraien:
`unstable` CI is expected to pass once the relevant PR's will be merge.

ephraimfeldblum (approved):
LGTM

PR #58, opened 2025-04-06
MOD-9192 Add buffer utils functionality

ephraimfeldblum:
[MOD-9192](https://redislabs.atlassian.net/browse/MOD-9192)

[MOD-9192]: https://redislabs.atlassian.net/browse/MOD-9192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/buffer.c
+++ b/src/utils/buffer.c
@@ -8,7 +8,6 @@
 #include "buffer.h"
 #include <string.h>
 #include "../mr_memory.h"
-#include "../mr.h"
 #include "../redismodule.h"
 
 void mr_BufferInit(mr_Buffer* buff, size_t initialCap){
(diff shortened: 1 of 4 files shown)


src/utils/buffer.c
```diff
@@ -94,3 +93,18 @@ char* mr_BufferReaderReadString(mr_BufferReader* br, int* error){
     return mr_BufferReaderReadBuff(br, &len, error);
 }
 
+void mr_BufferReaderRewind(mr_BufferReader* br) {
+    br->location = 0;
+}
+void mr_BufferWriterRewind(mr_BufferWriter* bw) {
+    bw->buff->size = 0;
+    bw->buff->cap = DEFAULT_INITIAL_CAP;
+    bw->buff->buff = MR_REALLOC(bw->buff->buff, DEFAULT_INITIAL_CAP);
```
galcohen-redislabs, line 102: Better to avoid reuse of the `DEFAULT_INITIAL_CAP` and use `bw->buff->cap` here instead.

Also, a question: who frees this buffer? Are there scenarios we get here with an already-allocated `buff` and can cause a leak?

ephraimfeldblum: yea, no need to realloc the buffer at all.

creation and destruction of buffers should be handled by libMR, the user doesn't need concern themselves with it. aside from that, i'm not sure i understand the question. realloc frees the old allocation if it determines to move to a larger allocation.

galcohen-redislabs: Yes, I misread it as `MR_ALLOC()`...

galcohen-redislabs (approved):

PR #92, opened 2026-05-19
MOD-15749 Add support for short form of CLUSTERSET

galcohen-redislabs:
This initial commit splits the command into two cases: long and short (which currently does nothing).
It also refactors the long form to help writing a clean version for the upcoming short form.

---

> [!NOTE]
> **High Risk**
> Changes cluster topology discovery and inter-shard CLUSTERSET propagation; short form depends on new RedisModule cluster slot APIs and misconfiguration could break cross-shard messaging.
> 
> **Overview**
> Adds a **compact `CLUSTERSET` form** (`CLUSTERSET` or `CLUSTERSET AUTH {pwd}`) alongside the existing verbose topology command, and refactors how cluster state is built from incoming arguments.
> 
> **Short form** no longer requires the full `HASHFUNC` / `RANGES` / `SHARD` payload: the module discovers masters, addresses, flags, and slot ranges from the Redis server via `GetClusterNodesList`, `GetClusterNodeInfo`, and the newly wired **`GetClusterNodeSlotRanges`** API (declared in `redismodule.h`). Optional `AUTH` applies one password to all shard connections.
> 
> **Long form** behavior is preserved but split into helpers (`InitClusterData`, `ParseShardEntry`, `SetClusterDataLongForm`). Shared init copies argv for later `CLUSTERSETFROMSHARD` propagation; **only long form** still substitutes the per-target `MYID` token when pushing topology on reconnect/hello retries.
> 
> Command entry points accept either form via `IsLongFormClusterSet` / `IsShortFormClusterSet` instead of a fixed minimum argc of 10.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4de0d20839352b6c78aee6cbab0f785589900c20. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -986,15 +986,15 @@ static void SetMyId(Cluster* cluster, RedisModuleString** argv, int argc){
 }
 
 static void InitClusterData(Cluster* cluster, RedisModuleString** argv, int argc){
-    clusterCtx.CurrCluster->clusterSetCommand = MR_ALLOC(sizeof(char*) * argc);
-    clusterCtx.CurrCluster->clusterSetCommandSize = argc;
-    clusterCtx.CurrCluster->clusterSetCommand[0] = MR_STRDUP(CLUSTER_SET_FROM_SHARD_COMMAND);
+    cluster->clusterSetCommand = MR_ALLOC(sizeof(char*) * argc);
+    cluster->clusterSetCommandSize = argc;
+    cluster->clusterSetCommand[0] = MR_STRDUP(CLUSTER_SET_FROM_SHARD_COMMAND);
 
-    GenerateRunId(clusterCtx.CurrCluster);
-    CopyClusterSetArgs(clusterCtx.CurrCluster, argv, argc);
-    SetMyId(clusterCtx.CurrCluster, argv, argc);
+    GenerateRunId(cluster);
+    CopyClusterSetArgs(cluster, argv, argc);
+    SetMyId(cluster, argv, argc);
 
-    clusterCtx.CurrCluster->nodes = mr_dictCreate(&mr_dictTypeHeapStrings, NULL);
+    cluster->nodes = mr_dictCreate(&mr_dictTypeHeapStrings, NULL);
 }
 
 #define INTERNAL_PASSWORD_MAX_SIZE 100
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1133,8 +1133,6 @@ static void SetClusterDataShortForm(RedisModuleString** argv, int argc){
 
         if (!(flags & REDISMODULE_NODE_MASTER)) continue;  // Skip replica nodes
 
-        RedisModule_Assert(MR_GetNode(nodeId) == NULL);
-
         RedisModuleSlotRangeArray *slots = RedisModule_GetClusterNodeSlotRanges(mr_staticCtx, nodeId);
         RedisModule_Assert(slots != NULL);
         long long minSlot = 0, maxSlot = -1;  // min > max indicates a no-hslots range (i.e., used for slotless shards)
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1152,10 +1152,11 @@ static void SetClusterDataShortForm(RedisModuleString** argv, int argc){
             clusterCtx.maxSlot = maxSlot;
         }
 
-        for (size_t j = 1; j < slots->num_ranges; j++) {
+        for (size_t j = 0; j < slots->num_ranges; j++) {
             minSlot = slots->ranges[j].start;
             maxSlot = slots->ranges[j].end;
-            mr_listAddNodeTail(aMasterNode->slotRanges, NewSlotRange(minSlot, maxSlot));
+            if (j > 0)  // The 0 case is handled by the MR_CreateNode() above
+                mr_listAddNodeTail(aMasterNode->slotRanges, NewSlotRange(minSlot, maxSlot));
             for (uint16_t k = minSlot ; k <= maxSlot ; k++)
                 clusterCtx.CurrCluster->slots[k] = aMasterNode;
         }
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -699,7 +699,7 @@ static void MR_OnConnectCallback(const struct redisAsyncContext* c, int status){
 
     if(n->sendClusterTopologyOnNextConnect && clusterCtx.CurrCluster->clusterSetCommand){
         bool isLongForm = IsLongFormClusterSet(clusterCtx.CurrCluster->clusterSetCommandSize);
-        RedisModule_Log(mr_staticCtx, "notice", "Sending cluster (%s form) topology to %s (%s:%d) on after reconnect",
+        RedisModule_Log(mr_staticCtx, "notice", "Sending cluster (%s form) topology to %s (%s:%d) after reconnect",
                 isLongForm ? "long" : "short", n->id, n->ip, n->port);
         if (isLongForm)
             clusterCtx.CurrCluster->clusterSetCommand[CLUSTERSET_MYID_LONG_FORM_INDEX] = MR_STRDUP(n->id);
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -51,7 +51,7 @@
  *
  *   or short form (new, more compact; the module will get the topology from the server directly):
  *
- * <module-name>.CLUSTERSET AUTH {auth}  // the assumption is that all nodes use the same auth
+ * <module-name>.CLUSTERSET [AUTH {pwd}]  // the assumption is that all nodes use the same (or no) password
  */
 
 #define CLUSTERSET_MYID_LONG_FORM_INDEX  6
@@ -61,7 +61,7 @@ static bool IsLongFormClusterSet(int argc) {
 }
 
 static bool IsShortFormClusterSet(int argc) {
-    return argc == 3;
+    return argc == 1 || argc == 3;
 }
 
 #define CLUSTER_INNER_COMMUNICATION_COMMAND xstr(MODULE_NAME)".INNERCOMMUNICATION"
@@ -1099,11 +1099,20 @@ static void SetClusterDataShortForm(RedisModuleString** argv, int argc){
     InitClusterData(clusterCtx.CurrCluster, argv, argc);
     memcpy(clusterCtx.myId, clusterCtx.CurrCluster->myId, REDISMODULE_NODE_ID_LEN + 1);
 
-    size_t index = 1;
-    const char *token = RedisModule_StringPtrLen(argv[index], NULL);
-    RedisModule_Assert(strcasecmp(token, "AUTH") == 0);
-    index++;
-    const char *password = RedisModule_StringPtrLen(argv[index], NULL);
+    const char *password = NULL;
+    switch (argc) {
+    case 1:
+        password = NULL;
+        break;
+    case 3: {
+        const char *token = RedisModule_StringPtrLen(argv[1], NULL);
+        RedisModule_Assert(strcasecmp(token, "AUTH") == 0);
+        password = RedisModule_StringPtrLen(argv[2], NULL);
+        break;
+    }
+    default:
+        RedisModule_Assert(0);
+    }
 
     size_t numNodes;
     char **nodeList = RedisModule_GetClusterNodesList(mr_staticCtx, &numNodes);
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -57,7 +57,7 @@
  */
 
 #define CLUSTERSET_MYID_LONG_FORM_INDEX  6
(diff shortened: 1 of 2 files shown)


gabsow (commented):
Solid refactor overall — splitting the long-form into `ParseShardEntry` reads well, and the short-form path matches MOD-15749's intent. A few things to address, the first one blocking:

| Severity | Issue |
|---|---|
| 🔴 Blocking | `slots[]` routing table missing the first range of every node — see inline comment on line 1147 |
| 🟠 Important | Slotless fallback claims all 16384 slots |
| 🟠 Important | AUTH password sent even when internal secret is in use (inconsistent with `MR_RefreshClusterData`) |
| 🟡 Cleanup | `InitClusterData` ignores its parameter; `isMe` double-set; log typo; redundant assertion; `uint16_t` loop bound; argc-based form detection is brittle |
| 🟢 Test gap | Only the happy-path 3-shard / no-AUTH case is covered; no slot-routed assertion (the test would pass even with the first-range bug) |

src/cluster.c
```diff
@@ -31,10 +31,39 @@
...
+ * <module-name>.CLUSTERSET [AUTH {pwd}]  // the assumption is that all nodes use the same (or no) password
+ */
+
+#define CLUSTERSET_MYID_LONG_FORM_INDEX  6
+
+static bool IsLongFormClusterSet(int argc) {
+    return argc >= 10;
+}
+
+static bool IsShortFormClusterSet(int argc) {
```
gabsow, line 63: Argc-based form detection is a bit brittle: short form is `argc == 1 || argc == 3`, long form is `argc >= 10`. If the long form ever evolved to use exactly 3 args (or short form to a config with 10+), this would silently dispatch wrong. A leading-token sniff (`AUTH` / `HASHFUNC` / nothing) would be more robust. Not blocking, but worth a comment at least.

src/cluster.c
```diff
@@ -664,11 +698,16 @@ static void MR_OnConnectCallback(const struct redisAsyncContext* c, int status){
     SendAuthCommandIfNeeded(c, n);
 
     if(n->sendClusterTopologyOnNextConnect && clusterCtx.CurrCluster->clusterSetCommand){
-        RedisModule_Log(mr_staticCtx, "notice", "Sending cluster topology to %s (%s:%d) after reconnect", n->id, n->ip, n->port);
-        clusterCtx.CurrCluster->clusterSetCommand[CLUSTER_SET_MY_ID_INDEX] = MR_STRDUP(n->id);
+        bool isLongForm = IsLongFormClusterSet(clusterCtx.CurrCluster->clusterSetCommandSize);
+        RedisModule_Log(mr_staticCtx, "notice", "Sending cluster (%s form) topology to %s (%s:%d) on after reconnect",
```
gabsow: Typo: `"on after reconnect"` → `"after reconnect"` (extra `on`).

src/cluster.c
```diff
@@ -915,58 +954,236 @@ static void MR_RefreshClusterData(){
...
-        RedisModule_Log(mr_staticCtx, "warning", "Could not parse cluster set arguments");
-        return;
+    cluster->myId = MR_ALLOC(REDISMODULE_NODE_ID_LEN + 1);
+    size_t zerosPadding = REDISMODULE_NODE_ID_LEN - myIdLen;
+    memset(cluster->myId, '0', zerosPadding);
+    memcpy(cluster->myId + zerosPadding, myId, myIdLen);
+    cluster->myId[REDISMODULE_NODE_ID_LEN] = '\0';
+}
+
+static void InitClusterData(Cluster* cluster, RedisModuleString** argv, int argc){
```
gabsow, line 988: `InitClusterData` takes a `Cluster* cluster` parameter but ignores it — every line mutates `clusterCtx.CurrCluster` directly. Either use the parameter throughout (so the helper is testable in isolation and the global mutation is explicit at the call site) or drop the parameter and rename to make it obvious it operates on the global. `SetMyId` and `GenerateRunId` get this right.

src/cluster.c
```diff
@@ -915,58 +954,236 @@ static void MR_RefreshClusterData(){
...
+        if (RedisModule_GetClusterNodeInfo(mr_staticCtx, nodeId, ip, NULL, &port, &flags) != REDISMODULE_OK) {
+            RedisModule_Log(mr_staticCtx, "warning", "Failed to get info for node %s", nodeId);
             continue;
         }
-        const char* arg = RedisModule_StringPtrLen(argv[i], NULL);
-        clusterCtx.CurrCluster->clusterSetCommand[i] = MR_STRDUP(arg);
+
+        if (!(flags & REDISMODULE_NODE_MASTER)) continue;  // Skip replica nodes
+
+        RedisModule_Assert(MR_GetNode(nodeId) == NULL);
```
gabsow: Redundant: `MR_CreateNode` already asserts `!MR_GetNode(id)` (see line 831). Either drop this pre-check or drop the one inside `MR_CreateNode`.

src/cluster.c
```diff
@@ -915,58 +954,236 @@ static void MR_RefreshClusterData(){
...
+
+        RedisModuleSlotRangeArray *slots = RedisModule_GetClusterNodeSlotRanges(mr_staticCtx, nodeId);
+        RedisModule_Assert(slots != NULL);
+        uint16_t minSlot = 0;
+        uint16_t maxSlot = NUMBER_OF_SLOTS - 1;
```
gabsow: 🟠 Slotless-node fallback is incorrect. When `slots->num_ranges == 0` (per redis/redis#14953, this happens for unknown nodes / non-cluster mode and plausibly for slotless masters), we fall back to `minSlot=0, maxSlot=NUMBER_OF_SLOTS-1` — i.e. we declare this node owns *all* 16384 slots.

The long form represents slotless with `minSlot=0, maxSlot=-1` so that `MR_CreateNode` adds no range (see `if (minSlot <= maxSlot)` at line 835). The short form should match — pass `0, -1` (use `long long` like the long form) and skip the `clusterCtx.minSlot/maxSlot` overwrite when slotless.

src/cluster.c
```diff
@@ -915,58 +954,236 @@ static void MR_RefreshClusterData(){
...
+        if (slots->num_ranges > 0) {
+            minSlot = slots->ranges[0].start;
+            maxSlot = slots->ranges[0].end;
+        }
+
+        Node* aMasterNode = MR_CreateNode(nodeId, ip, port, password, NULL, minSlot, maxSlot);
```
gabsow, line 1144: 🔴 **Blocking bug: `clusterCtx.CurrCluster->slots[]` routing table is never populated for the first range.**

`MR_CreateNode` only stores the first range in the *node's own* `slotRanges` list (line 836) — it does **not** write into the global `slots[]` table. The loop below at line 1155 then iterates `j = 1` onward, so the first range of every node is missing from `slots[]`.

Concrete case: a normal 3-shard cluster where each shard owns one contiguous range → after short-form CLUSTERSET, `slots[]` is entirely NULL. The dispatch at line 283 (`slots[sendMsg->slotToSend]`) will then NULL-deref or silently misroute every slot-targeted message.

Compare with `SetClusterDataLongForm` (line 1220-1222) and `MR_RefreshClusterData` (line 948-950) — both unconditionally fill `slots[k]` for every range.

Fix: either run the slot-fill loop *before* the `j=1..` extra-range loop, or unify both into a single `for (j=0; j<num_ranges; j++)` block:

```c
Node* aMasterNode = MR_CreateNode(nodeId, ip, port, password, NULL, 0, -1);
aMasterNode->isMe = (flags & REDISMODULE_NODE_MYSELF) != 0;
for (size_t j = 0; j < slots->num_ranges; j++) {
    uint16_t minSlot = slots->ranges[j].start;
    uint16_t maxSlot = slots->ranges[j].end;
    mr_listAddNodeTail(aMasterNode->slotRanges, NewSlotRange(minSlot, maxSlot));
    for (int k = minSlot; k <= maxSlot; k++)
        clusterCtx.CurrCluster->slots[k] = aMasterNode;
    if (aMasterNode->isMe && j == 0) {
        clusterCtx.minSlot = minSlot;
        clusterCtx.maxSlot = maxSlot;
    }
}
```

Note: the test `test_short_form_clusterset` did NOT catch this because `TS.QUERYINDEX` fans out via the *node dict*, not via `slots[]`. The dict is populated correctly, so the broadcast works. See my comment on the RTS PR for a suggested slot-routed test.

src/cluster.c
```diff
@@ -915,58 +954,236 @@ static void MR_RefreshClusterData(){
...
+        aMasterNode->isMe = (flags & REDISMODULE_NODE_MYSELF) != 0;
```
gabsow, line 1150: `isMe` is set twice — `MR_CreateNode` sets it via `strcmp(id, myId)` at line 862, then this line overrides via the `MYSELF` flag. They should always agree, but having two sources of truth is confusing. The `MYSELF` flag is the more direct signal — consider making `MR_CreateNode` not set `isMe` at all when called from this path, or unify.

src/cluster.c
gabsow, line 1144: 🟠 The password obtained from `AUTH <pwd>` is passed unconditionally to `MR_CreateNode`. Compare with `MR_RefreshClusterData` (line 940):
```c
n = MR_CreateNode(..., RedisModule_GetInternalSecret ? NULL : clusterCtx.password, NULL, ...);
```
i.e. when an internal secret is available, password auth is intentionally skipped. If a deployment uses internal secrets, the short form will still attempt password auth — at minimum inconsistent with the existing pattern, possibly wrong. Either mirror the same guard or leave a comment explaining the difference.

src/cluster.c
```diff
@@ -915,58 +954,236 @@ static void MR_RefreshClusterData(){
...
+            clusterCtx.minSlot = minSlot;
+            clusterCtx.maxSlot = maxSlot;
+        }
+
+        for (size_t j = 1; j < slots->num_ranges; j++) {
+            minSlot = slots->ranges[j].start;
+            maxSlot = slots->ranges[j].end;
+            mr_listAddNodeTail(aMasterNode->slotRanges, NewSlotRange(minSlot, maxSlot));
+            for (uint16_t k = minSlot ; k <= maxSlot ; k++)
+                clusterCtx.CurrCluster->slots[k] = aMasterNode;
```
gabsow, line 1163: `uint16_t k` is safe for canonical `maxSlot ≤ 16383`, but would spin forever if a buggy API ever returns `maxSlot == UINT16_MAX` (e.g. casting `-1`). The long-form loop at line 1220 uses `int k`. Match that here.

src/cluster.c
galcohen-redislabs: They will not change that much. Actually we plan to get rid of both. So argc-based logic is good enough.

src/cluster.c
galcohen-redislabs: oy! fixed.

src/cluster.c
galcohen-redislabs: dropped

src/cluster.c
galcohen-redislabs: fixed

src/cluster.c
galcohen-redislabs: good catch. fixed.
(we don't use the `slots[]` anyway in ts because it never routes via a hash. the cluster-wide commands always send to all nodes)

galcohen-redislabs: Note that this is not the same case. Here we fetch the password from the parsing the CLUSTERSET command (similar to what's being done in the parsing of ADDR in the long form). The CLUSTERSET from the DMC will always have a password (which is the bdb's `internal_pass` attribute).

src/cluster.c
galcohen-redislabs: This is true, and the flag is the more "correct" way. But note that MR_CreateNode is also called from other' flag-less' environment, so we need the strcmp logic as well. I'll add a comment.

src/cluster.c
galcohen-redislabs: matched

gabsow (approved):
LGTM ✅

All 9 inline comments addressed:
- 🔴 Blocking `slots[]` first-range bug → fixed in `42ce87e`
- 🟠 Slotless fallback → fixed in `458f2ca` (now uses `long long 0, -1` sentinel)
- 🟠 AUTH vs internal-secret → accepting your reasoning (symmetric with long-form ADDR parsing; DMC always provides the password)
- 🟡 `InitClusterData` parameter, redundant assert, log typo, `uint16_t`→`int` loop, `isMe` rationale comment → all addressed
- 🟡 Argc-based form detection → accepting (you plan to retire both forms)

Flagging for downstream visibility: the RTS PR (#2034) still pins to `ad6a45d3` and CI is fully red because the blocking `slots[]` bug surfaces in tests there. Bumping the submodule to this PR's HEAD (or to the merge commit) should unblock it.

PR #93, opened 2026-05-26
MOD-15862 Diagnostic: bound SSL CI runs with per-test + step timeouts

gabsow:
## Summary
- Linux CI runs were being cancelled at the 6h job cap. Root cause: commit [2857cfe](https://github.com/RedisGears/LibMR/commit/2857cfec3c78e66f2548f533becb96df640cb56b) (MOD-14850) added a single-shard `--tls` invocation to `make test_ssl` so the new `testNoCrashOnTLSHandshakeFailure` could run. That invocation also un-skipped the entire `test_network.py` ShardMock suite under TLS for the first time, and at least one of those tests is hanging on Linux.
- The hang is invisible in the log because `run_tests.sh` passes RLTest `--no-progress` and the default `--test-timeout 0` (no timeout), so the test name never makes it out before the job is killed.

This PR is the diagnostic step — not the fix:
- Drop `--no-progress` from `run_tests.sh` and add `--test-timeout 120`, so any single test that hangs gets killed at 2 min and shows up by name.
- Add `timeout-minutes: 45` to the Linux job and `timeout-minutes: 25` to the SSL step so a wedged run fails fast instead of burning 6h.

Once this run completes, the failing test name will be in the log and the next PR can fix the actual hang.

## Test plan
- [ ] Linux CI run finishes within 45 minutes (vs. previous 6h hang)
- [ ] If a test still hangs, the offending test name is visible in the SSL step output before it's killed by `--test-timeout`
- [ ] macOS CI continues to pass on `unstable`/`7.4` (no behaviour change expected there)

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI configuration and test harness behavior; no production cluster or auth logic is modified.
> 
> **Overview**
> Adds **CI guardrails** so wedged Linux SSL runs fail fast instead of hitting the 6h job cap: **45m** job timeout on Linux/macOS, **25m** on the SSL test step, and on Linux SSL **`PYTHONUNBUFFERED`**, plus **`RLTEST_EXTRA_ARGS`** (e.g. `--test-timeout 300`) passed through **`run_tests.sh`**.
> 
> Relaxes **`testSendRetriesMechanizm`** so it no longer assumes exactly three hard-coded error/reconnect cycles. It now loops with timeouts, allows **1–3** `INNERCOMMUNICATION` attempts (TLS can queue the first send in the retry path), and still asserts **no reconnect** after `MSG_MAX_RETRIES`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 09446a25af6ff1e234f39eb91e8c1067004f4549. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -66,7 +66,7 @@ jobs:
       env:
         PYTHON: python
         PYTHONUNBUFFERED: "1"
-        RLTEST_EXTRA_ARGS: "--test-timeout 180"
+        RLTEST_EXTRA_ARGS: "--test-timeout 300"
     - name: Valgrind tests
       run: make run_tests_valgrind
       env:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -13,6 +13,7 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    timeout-minutes: 45
 
     strategy:
       fail-fast: false
@@ -60,6 +61,7 @@ jobs:
       env:
         PYTHON: python
     - name: SSL tests
+      timeout-minutes: 25
       run: make run_tests_ssl
       env:
         PYTHON: python
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -65,6 +65,8 @@ jobs:
       run: make run_tests_ssl
       env:
         PYTHON: python
+        PYTHONUNBUFFERED: "1"
+        RLTEST_EXTRA_ARGS: "--test-timeout 180"
     - name: Valgrind tests
       run: make run_tests_valgrind
       env:
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -15,6 +15,7 @@ jobs:
   build:
 
     runs-on: macos-latest
+    timeout-minutes: 45
 
     strategy:
       fail-fast: false
@@ -51,6 +52,7 @@ jobs:
       env:
         PYTHON: python
     - name: SSL tests
+      timeout-minutes: 25
       run: make run_tests_ssl
       env:
         PYTHON: python
--- a/tests/mr_test_module/pytests/run_tests.sh
+++ b/tests/mr_test_module/pytests/run_tests.sh
@@ -19,4 +19,4 @@ else
 fi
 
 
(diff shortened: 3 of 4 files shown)


gabsow:
## Root cause found and fixed in latest commit ([175346e](https://github.com/RedisGears/LibMR/pull/93/commits/175346e))

The hung test is **`testSendRetriesMechanizm`** in [test_network.py:432](https://github.com/RedisGears/LibMR/blob/175346e/tests/mr_test_module/pytests/test_network.py#L432). It was hardcoded to send 3 `-Err` responses and read 3 `MRTESTS.INNERCOMMUNICATION` requests — but under TLS, libmr only sends **two** before giving up.

### Why the count differs

Per [src/cluster.c:445-451](https://github.com/RedisGears/LibMR/blob/175346e/src/cluster.c#L445), `MSG_MAX_RETRIES=3` and the resend loop increments `retries` *before* checking `retries < 3`:

- **Non-TLS:** `NETWORKTEST` runs after `NodeStatus_Connected`, so `MR_ClusterSendMsgToNode` sends synchronously (`retries=0`). After each `-Err` + reconnect, `MR_HelloResponseArrived` re-sends and bumps retries to 1, 2, 3. Total **3 sends**.
- **TLS:** the TLS+AUTH+HELLO handshake is slower, so `NETWORKTEST` runs while status is `NodeStatus_HelloSent`. The msg hits the "message was not sent because status is not connected" path, gets queued, and the actual first send happens via the resend loop — which burns retry #1. Only **2 sends** total.

The redis log embedded in the timeout dump confirmed this: only two `Received an invalid status reply` log lines before `Gave up of message because failed to send it for more than 3 time`.

### Why this surfaced now

Commit [2857cfe](https://github.com/RedisGears/LibMR/commit/2857cfe) (MOD-14850) added a single-shard `--tls` invocation to `test_ssl`. `testSendRetriesMechanizm` has `skipOnCluster=True` and was previously only ever run in cluster mode under TLS — i.e., always skipped. Under the new single-shard `--tls` invocation it ran for the first time and immediately hung. macOS happens to pass it (probably faster TLS path → status reaches Connected before NETWORKTEST), so only Linux saw the 6h burn.

### Fix

Rewrite the test to express the actual invariant — libmr sends the message between 1 and `MSG_MAX_RETRIES` times then stops — with bounded read/GetConnection timeouts so the test cannot hang regardless of where the count lands.

PR #97, opened 2026-06-10
Drain LibMR threads to a safe point before fork() (MOD-15307)

gabsow:
## Summary
Adds `MR_DrainForFork()` / `MR_ResumeAfterFork()`. Called from the embedding module's pre-fork handler (main thread), `MR_DrainForFork()`:
- parks the event-loop thread at a **between-tasks** safe point via a posted task, and
- **bounded-waits** the worker pool to idle,

so no LibMR thread holds a libc lock at `fork()` (which would ghost-lock the child). `MR_ResumeAfterFork()` releases the parked event-loop thread (called after the fork, on success or cancel).

Cooperative, bounded, fail-open. Deliberately **not** the existing `mr_thpool_pause` (SIGUSR2), which can freeze a worker mid-`malloc` holding the arena lock — exactly the ghost-lock this prevents. A worker still blocked acquiring the GIL is already malloc-safe, so a drain timeout there is benign.

## Why / depends on
Fixes the RedisTimeSeries ASM-migration nightly hangs (MOD-14615 valgrind, MOD-14239 sanitizer). The embedding module wires this to redis core's new `FORK_CHILD_PRE` subevent (redis/redis#15327).

## Pre-merge note
The two `RedisModule_Log(..., "notice", ...)` lines in `MR_DrainForFork`/`MR_ResumeAfterFork` should be downgraded to `debug` (kept at `notice` to confirm the drain fires during CI validation).

---

> [!NOTE]
> **Medium Risk**
> Fork-time threading and allocator safety are sensitive; incomplete drain can still leave ghost locks, though behavior is bounded and fail-open with warnings.
> 
> **Overview**
> Adds **`MR_DrainForFork()`** and **`MR_ResumeAfterFork()`** so embedding modules can quiesce LibMR immediately before `fork()`. **`MR_DrainForFork()`** posts a task that parks the event-loop thread at a between-tasks safe point, then **bounded-waits** (2s) for that park and for the execution worker pool to go idle via new **`mr_thpool_wait_timeout()`**. If quiescence is incomplete, it logs a warning and continues (**fail-open**). **`MR_ResumeAfterFork()`** unblocks the parked event-loop thread and must be paired with every drain (after fork success or cancel).
> 
> This is intentionally **not** `mr_thpool_pause` (SIGUSR2), which can stop workers mid-`malloc` and worsen ghost-lock risk.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df7cc0902ab672e4722f01f8ee3af1cb2d2da8d0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1731,6 +1731,31 @@ size_t MR_ClusterGetSize(){
     return clusterCtx.clusterSize;
 }
 
+/* MOD-14615: format the peer shards that are NOT me and NOT present in the
+ * `responded` set into "id(ip:port),..." in `out`, so a max-idle timeout can
+ * name the non-responding shard(s). Runs on the event-loop thread (same as the
+ * message handlers), so reading the cluster node table needs no extra locking.
+ * The node-id is the same NUL-terminated string used as the responded-set key. */
+void MR_ClusterFormatPendingPeers(mr_dict* responded, char* out, size_t outLen) {
+    if (outLen == 0) return;
+    out[0] = '\0';
+    if (!clusterCtx.CurrCluster || !clusterCtx.CurrCluster->nodes) return;
+    mr_dictIterator *iter = mr_dictGetIterator(clusterCtx.CurrCluster->nodes);
+    mr_dictEntry *entry = NULL;
+    size_t off = 0;
+    while ((entry = mr_dictNext(iter))) {
+        Node* n = mr_dictGetVal(entry);
+        if (n->isMe) continue;
+        if (responded && mr_dictFind(responded, n->id)) continue;
+        int w = snprintf(out + off, outLen - off, "%s%s(%s:%u)",
+                         off ? "," : "", n->id, n->ip ? n->ip : "?", n->port);
+        if (w < 0) break;
+        if ((size_t)w >= outLen - off) { out[outLen - 1] = '\0'; break; } /* truncated */
+        off += (size_t)w;
+    }
+    mr_dictReleaseIterator(iter);
+}
+
 int MR_ClusterIsClusterMode(){
     return MR_ClusterGetSize() > 1;
 }
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1800,31 +1800,6 @@ size_t MR_ClusterGetSize(){
     return clusterCtx.clusterSize;
 }
 
-/* MOD-14615: format the peer shards that are NOT me and NOT present in the
- * `responded` set into "id(ip:port),..." in `out`, so a max-idle timeout can
- * name the non-responding shard(s). Runs on the event-loop thread (same as the
- * message handlers), so reading the cluster node table needs no extra locking.
- * The node-id is the same NUL-terminated string used as the responded-set key. */
-void MR_ClusterFormatPendingPeers(mr_dict* responded, char* out, size_t outLen) {
-    if (outLen == 0) return;
-    out[0] = '\0';
-    if (!clusterCtx.CurrCluster || !clusterCtx.CurrCluster->nodes) return;
-    mr_dictIterator *iter = mr_dictGetIterator(clusterCtx.CurrCluster->nodes);
-    mr_dictEntry *entry = NULL;
-    size_t off = 0;
-    while ((entry = mr_dictNext(iter))) {
-        Node* n = mr_dictGetVal(entry);
-        if (n->isMe) continue;
-        if (responded && mr_dictFind(responded, n->id)) continue;
-        int w = snprintf(out + off, outLen - off, "%s%s(%s:%u)",
-                         off ? "," : "", n->id, n->ip ? n->ip : "?", n->port);
-        if (w < 0) break;
-        if ((size_t)w >= outLen - off) { out[outLen - 1] = '\0'; break; } /* truncated */
-        off += (size_t)w;
-    }
-    mr_dictReleaseIterator(iter);
-}
-
 int MR_ClusterIsClusterMode(){
     return MR_ClusterGetSize() > 1;
 }
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -172,6 +172,7 @@ typedef struct ClusterSetCtx {
     RedisModuleString **argv;
     int argc;
     bool force;
+    const char* errReply;  // NULL => reply "OK"; otherwise reply this error to the client
 }ClusterSetCtx;
 
 typedef enum MessageReply {
@@ -415,7 +416,7 @@ static void SendAuthCommandIfNeeded(const struct redisAsyncContext* c, const Nod
         redisAsyncCommand((redisAsyncContext*)c, NULL, NULL, "AUTH %s", n->password);
         return;
     }
-    if (RedisModule_GetInternalSecret && !MR_IsEnterpriseBuild()) {
+    if (RedisModule_GetInternalSecret && clusterCtx.isOss) {
         /* OSS deployment that support internal secret, lets use it. */
         RedisModule_ThreadSafeContextLock(mr_staticCtx);
         size_t len;
@@ -449,7 +450,7 @@ static void MR_HelloResponseArrived(struct redisAsyncContext* c, void* a, void* 
             // accepted connections but did not set its internal secret to the cluster's yet.
             // In such cases we will have a regular (i.e., non-internal) connection and the
             // hidden commands (including `<module>.HELLO`) will not be visible to us.
-            if (!MR_IsEnterpriseBuild() && strstr(reply->str, "unknown command") != NULL)
+            if (clusterCtx.isOss && strstr(reply->str, "unknown command") != NULL)
                 SendAuthCommandIfNeeded(c, n);
         }
         n->resendHelloEvent = MR_EventLoopAddTaskWithDelay(MR_ClusterResendHelloMessage, n, RETRY_INTERVAL);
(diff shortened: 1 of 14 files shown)


src/mr.c
```diff
@@ -1408,6 +1468,121 @@ void MR_ExecutionSetMaxIdle(Execution* e, size_t maxIdle) {
...
+    while (mr_thpool_num_threads_working(mrCtx.executionsThreadPool) > 0) {
+        if (mr_forkDeadlinePassed(&deadline)) {
+            RedisModule_Log(mr_staticCtx, "warning",
+                "MR_DrainForFork: %d worker(s) still busy at timeout; forking anyway",
+                mr_thpool_num_threads_working(mrCtx.executionsThreadPool));
+            break;
+        }
+        struct timespec nap = { .tv_sec = 0, .tv_nsec = 1000000L }; /* 1ms */
+        nanosleep(&nap, NULL);
+    }
```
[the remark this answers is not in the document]
gabsow: Fixed in f6fe687. `MR_DrainForFork` now waits for the pool to be fully idle — both in-flight workers **and** queued-but-not-yet-started jobs — via a new `mr_thpool_num_jobs_in_queue()` accessor:

```c
while (mr_thpool_num_threads_working(pool) > 0 ||
       mr_thpool_num_jobs_in_queue(pool) > 0) { ... }
```

This matches the invariant `thpool_wait()` already uses (`while (jobqueue.len || num_threads_working)`). Since the event-loop thread is parked first, no new work is dispatched, so once both reach 0 the pool stays idle through the fork. Good catch.

src/utils/thpool.c
```diff
@@ -314,6 +314,10 @@ int mr_thpool_num_threads_working(mr_thpool_* thpool_p) {
   return thpool_p->num_threads_working;
 }
 
+int mr_thpool_num_jobs_in_queue(mr_thpool_* thpool_p) {
+  return thpool_p->jobqueue.len;
+}
```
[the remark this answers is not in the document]
gabsow: Fixed in 995f842 — `mr_thpool_num_jobs_in_queue()` now reads `jobqueue.len` under `jobqueue.rwmutex` (the same lock `jobqueue_push`/`jobqueue_pull` mutate it under), so the drain no longer races on a stale length.

src/mr.c
```diff
@@ -1408,6 +1468,126 @@ void MR_ExecutionSetMaxIdle(Execution* e, size_t maxIdle) {
...
+        if (mr_forkDeadlinePassed(&deadline)) {
+            RedisModule_Log(mr_staticCtx, "warning",
+                "MR_DrainForFork: pool not idle at timeout (busy=%d, queued=%d); forking anyway",
+                mr_thpool_num_threads_working(mrCtx.executionsThreadPool),
+                mr_thpool_num_jobs_in_queue(mrCtx.executionsThreadPool));
+            break;
+        }
+        struct timespec nap = { .tv_sec = 0, .tv_nsec = 1000000L }; /* 1ms */
+        nanosleep(&nap, NULL);
+    }
```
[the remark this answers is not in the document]
gabsow: This one is intentional. `MR_DrainForFork` runs synchronously on redis's main thread before every fork, so the single `MR_FORK_DRAIN_TIMEOUT_MS` (2s) deadline deliberately bounds the **total** main-thread stall — we do not want a fork (RDB/AOF/ASM snapshot) to block the server longer than that.

In the normal case the park task runs in microseconds, so phase two effectively gets the full budget (measured median drain ~62us). If the park itself cannot complete within the budget, the event-loop thread is stuck off a safe point and the correct bounded fallback is to fork anyway — the drain is best-effort, not a hard barrier. Giving phase two its own fresh 2s budget would double the worst-case main-thread stall, which we deliberately avoid.

gabsow:
@galcohen-redislabs wanted your opinion on the last commit (`995f842`) before we lean on it. It takes `jobqueue.rwmutex` when reading `jobqueue.len` in `mr_thpool_num_jobs_in_queue`, to address the Bugbot "unlocked job-queue length read" finding.

Two things I am not sure about:

1. **Is it really necessary?** The neighbouring `mr_thpool_num_threads_working()` reads `num_threads_working` unlocked as well, and `MR_DrainForFork` polls both in a 1ms loop that is best-effort (it forks anyway on timeout) — so a stale `len` self-corrects on the next tick. By that logic the unlocked read is arguably benign (same as the existing one), and we could just reply to the bot instead of adding a lock.

2. **Isn't it dangerous?** It adds a lock acquisition on the **main thread, in the pre-fork drain path**. As far as I can tell it's safe — the lock is taken and released inside the accessor (never held across `fork()`), and workers only hold `rwmutex` briefly in `jobqueue_push`/`jobqueue_pull` without needing the GIL, so there's no GIL/lock inversion or fork-time deadlock. But I would rather get your read before relying on it.

Happy to drop the lock and instead annotate the read as a benign race (matching `num_threads_working`) if you think that is cleaner/safer. What do you think?

src/cluster.c
```diff
@@ -1800,6 +1800,31 @@ size_t MR_ClusterGetSize(){
     return clusterCtx.clusterSize;
 }
 
+/* MOD-14615: format the peer shards that are NOT me and NOT present in the
+ * `responded` set into "id(ip:port),..." in `out`, so a max-idle timeout can
+ * name the non-responding shard(s). Runs on the event-loop thread (same as the
+ * message handlers), so reading the cluster node table needs no extra locking.
+ * The node-id is the same NUL-terminated string used as the responded-set key. */
```
galcohen-redislabs: The whole comment is too specific: no need to specify the ticket number and no need to explain the implementation and cause for this specific issue.

src/cluster.c
```diff
@@ -1800,6 +1800,31 @@ size_t MR_ClusterGetSize(){
...
+    out[0] = '\0';
+    if (!clusterCtx.CurrCluster || !clusterCtx.CurrCluster->nodes) return;
+    mr_dictIterator *iter = mr_dictGetIterator(clusterCtx.CurrCluster->nodes);
+    mr_dictEntry *entry = NULL;
+    size_t off = 0;
+    while ((entry = mr_dictNext(iter))) {
+        Node* n = mr_dictGetVal(entry);
+        if (n->isMe) continue;
+        if (responded && mr_dictFind(responded, n->id)) continue;
+        int w = snprintf(out + off, outLen - off, "%s%s(%s:%u)",
```
galcohen-redislabs: The size_t math is dangerous. As a rule of thumb you don't want to subtract two unsigned (for the fear of a wrap-around to large numbers).
Better to add an explicit guard here and change the comparison below to use additions only.

src/cluster.h
```diff
@@ -35,6 +35,13 @@ int MR_IsClusterInitialize();
 
 size_t MR_ClusterGetSize();
 
+/* MOD-14615: format peer shards (id(ip:port)) that are NOT in `responded` into
```
galcohen-redislabs: Here too: no need to include ticket number or implementation details or reasoning for doing so or apologies about the struct declaration...

galcohen-redislabs: (also in all other places; we don't need comments explaining how and why code was written to solve a specific ticket)

src/cluster.c
galcohen-redislabs: The new diagnostic data of more explicit names (of workers that didn't respond after a timeout, etc.), including the needed locking mechanism and the monotonic clock, etc. are not really related to the core issue in this ticket (i.e., the forking coordination between the module and the redis core).

The whole thing is just a code slop and should not be included in this PR. If we really want such code (only if it is really a big issue, which honestly I don't think it is...) then we should open a specific ticket for that.

src/mr.c
```diff
@@ -1408,6 +1468,126 @@ void MR_ExecutionSetMaxIdle(Execution* e, size_t maxIdle) {
...
+static int mr_forkElParked = 0;  /* set by the park task once the el-thread is quiesced */
+static int mr_forkElRelease = 0; /* set by MR_ResumeAfterFork to release the el-thread */
+
+/* Runs on the event-loop thread, between tasks (a safe point). Parks the thread until
+ * MR_ResumeAfterFork() is called. */
+static void MR_ForkParkElThread(void* ctx) {
+    REDISMODULE_NOT_USED(ctx);
+    pthread_mutex_lock(&mr_forkDrainLock);
+    mr_forkElParked = 1;
+    pthread_cond_broadcast(&mr_forkDrainCond);
```
galcohen-redislabs, line 1430: Why not use the already-existing sync. functions in utils?

galcohen-redislabs: Also: the broadcasting back and forth between ml and worker threads (with its extra locks and hidden locks such that in `mr_thpool_num_jobs_in_queue()`) pause a threat of increasing latencies for all commands. I would really like a much simpler solution, if possible.

gabsow:
@galcohen-redislabs thanks for the thorough review — addressed all of it in `df7cc09` (net −177/+42):

- **Removed the max-idle diagnostics entirely** — the responded-sets, monotonic timestamps, dispatch/timeout logging, `MR_ClusterFormatPendingPeers`, and `mr_thpool_num_jobs_in_queue`. Agreed it's out of scope for the fork-coordination fix; tracked separately as **MOD-16404** (your latency + `size_t` notes are captured there, and the original impl is referenced for whoever picks it up). Removing it also takes the `size_t` subtraction you flagged with it.
- **Simplified the drain to reuse existing sync** — the worker wait is now a new `mr_thpool_wait_timeout()` that simply bounds the existing `mr_thpool_wait` on the pool's own `thcount_lock`/`threads_all_idle`. No hand-rolled poll loop, no per-read `jobqueue` lock, and no el↔worker broadcasting on the command path — so no added per-command latency. The whole drain is now: park the event-loop thread, then `mr_thpool_wait_timeout(pool, 2s)`.
- **Trimmed the comments** — dropped ticket numbers and the how/why-for-this-ticket explanations throughout.

PTAL.

PR #98, opened 2026-06-11
MOD-15896 Detect oss cluster at MR_ClusterInit

AvivDavid23:
> [!NOTE]
> **Medium Risk**
> Changes load-time cluster classification and inter-shard AUTH/command ACLs; misreading `cluster-enabled` is logged and defaults conservatively, but wrong classification would break shard connectivity or enterprise proxy behavior.
> 
> **Overview**
> **Cluster behavior no longer keys off “enterprise build” alone.** At `MR_ClusterInit`, the module reads `CONFIG GET cluster-enabled` (handling RESP2 array and RESP3 map replies) and combines that with **`MR_RlecVersionPresent`** (whether `rlec_version` appeared in server INFO) to set **`clusterCtx.isOss`**. A shard is treated as enterprise only when `rlec_version` is present **and** cluster mode is not OSS—so an enterprise binary running as an OSS cluster (e.g. env0 tests) uses the OSS path: internal-secret AUTH, `internal` command flags, and no `_proxy-filtered`.
> 
> **`MR_IsEnterpriseBuild()` / `MR_IsEnterprise` are removed.** Version parsing in `MR_GetRedisVersion` now sets **`MR_RlecVersionPresent`** independently of whether version numbers parse, avoiding misclassifying unparseable enterprise strings as OSS. Hello/auth retry logic and command registration use **`clusterCtx.isOss`** instead of the old enterprise-build helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ebd3cf39b32e980928a3865b81385ba3e3c5246. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1659,10 +1659,12 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
     if (ceReply && RedisModule_CallReplyType(ceReply) == REDISMODULE_REPLY_ARRAY &&
         RedisModule_CallReplyLength(ceReply) >= 2) {
         RedisModuleCallReply *val = RedisModule_CallReplyArrayElement(ceReply, 1);
-        size_t vlen = 0;
-        const char *vstr = RedisModule_CallReplyStringPtr(val, &vlen);
-        if (vstr && vlen == 3 && strncmp(vstr, "yes", 3) == 0) {
-            ossClusterRuntime = true;
+        if (RedisModule_CallReplyType(val) == REDISMODULE_REPLY_STRING) {
+            size_t vlen = 0;
+            const char *vstr = RedisModule_CallReplyStringPtr(val, &vlen);
+            if (vstr && vlen == 3 && strncmp(vstr, "yes", 3) == 0) {
+                ossClusterRuntime = true;
+            }
         }
     }
     if (ceReply) {
@@ -1673,7 +1675,9 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
     }
     RedisModule_FreeServerInfo(rctx, info);
 
-    RedisModule_Log(rctx, "notice", "Detected redis %s", clusterCtx.isOss? "oss" : "enterprise");
+    RedisModule_Log(rctx, "notice", "Detected redis %s (cluster-enabled=%s)",
+                    clusterCtx.isOss ? "oss" : "enterprise",
+                    ossClusterRuntime ? "yes" : "no");
 
     const char *command_flags = "readonly deny-script";
     if (!clusterCtx.isOss) {
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1649,31 +1649,48 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
     clusterCtx.password = password ? MR_STRDUP(password) : NULL;
     memset(clusterCtx.myId, '0', REDISMODULE_NODE_ID_LEN);
 
-    RedisModuleServerInfoData *info = RedisModule_GetServerInfo(rctx, "Server");
-    const char *rlecVersion = RedisModule_ServerInfoGetFieldC(info, "rlec_version");
     /* Note: RedisModule_GetContextFlags() does NOT report
      * REDISMODULE_CTX_FLAGS_CLUSTER yet at module-load time, so read the
      * cluster-enabled config directly to learn the runtime mode. */
     bool ossClusterRuntime = false;
     RedisModuleCallReply *ceReply = RedisModule_Call(rctx, "config", "cc", "get", "cluster-enabled");
-    if (ceReply && RedisModule_CallReplyType(ceReply) == REDISMODULE_REPLY_ARRAY &&
-        RedisModule_CallReplyLength(ceReply) >= 2) {
-        RedisModuleCallReply *val = RedisModule_CallReplyArrayElement(ceReply, 1);
-        if (RedisModule_CallReplyType(val) == REDISMODULE_REPLY_STRING) {
+    if (ceReply) {
+        /* CONFIG GET replies as a flat array in RESP2 and as a map in RESP3.
+         * The module ctx defaults to RESP2, but handle both shapes so a future
+         * RESP3 default can't silently break detection and misclassify an
+         * OSS-cluster shard as enterprise. */
+        int ceType = RedisModule_CallReplyType(ceReply);
+        RedisModuleCallReply *val = NULL;
+        if (ceType == REDISMODULE_REPLY_ARRAY && RedisModule_CallReplyLength(ceReply) >= 2) {
+            val = RedisModule_CallReplyArrayElement(ceReply, 1);
+        } else if (ceType == REDISMODULE_REPLY_MAP && RedisModule_CallReplyMapElement &&
+                   RedisModule_CallReplyLength(ceReply) >= 1) {
+            RedisModule_CallReplyMapElement(ceReply, 0, NULL, &val);
+        }
+        if (val && RedisModule_CallReplyType(val) == REDISMODULE_REPLY_STRING) {
             size_t vlen = 0;
             const char *vstr = RedisModule_CallReplyStringPtr(val, &vlen);
-            if (vstr && vlen == 3 && strncmp(vstr, "yes", 3) == 0) {
+            /* CONFIG GET returns the canonical lowercase "yes"/"no" (Redis
+             * spec), so a case-sensitive compare of exactly 3 bytes is safe. */
+            if (vstr && vlen == 3 && memcmp(vstr, "yes", 3) == 0) {
                 ossClusterRuntime = true;
             }
         }
-    }
-    if (ceReply) {
         RedisModule_FreeCallReply(ceReply);
     }
-    if (rlecVersion && !ossClusterRuntime) {
+    /* rlec_version is only present on enterprise binaries. It was already
+     * parsed into MR_RlecMajorVersion by MR_GetRedisVersion() (run before
+     * MR_ClusterInit), so reuse it here instead of fetching Server info a
+     * second time; MR_RlecMajorVersion >= 0 means the field was present.
+     *
+     * Treat the shard as enterprise only when rlec_version is present AND we
+     * are not in OSS cluster mode. This is the load-bearing decision of the
+     * PR: it unblocks an enterprise binary running as an OSS cluster (e.g. the
+     * env0 testing setup), which must take the OSS path (internal-secret AUTH,
+     * no _proxy-filtered command flags) rather than the enterprise path. */
+    if (MR_RlecMajorVersion >= 0 && !ossClusterRuntime) {
         clusterCtx.isOss = false;
     }
-    RedisModule_FreeServerInfo(rctx, info);
 
     RedisModule_Log(rctx, "notice", "Detected redis %s (cluster-enabled=%s)",
                     clusterCtx.isOss ? "oss" : "enterprise",
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1653,6 +1653,7 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
      * REDISMODULE_CTX_FLAGS_CLUSTER yet at module-load time, so read the
      * cluster-enabled config directly to learn the runtime mode. */
     bool ossClusterRuntime = false;
+    bool ceResolved = false;
     RedisModuleCallReply *ceReply = RedisModule_Call(rctx, "config", "cc", "get", "cluster-enabled");
     if (ceReply) {
         /* CONFIG GET replies as a flat array in RESP2 and as a map in RESP3.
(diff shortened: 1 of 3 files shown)


src/cluster.c
```diff
@@ -1651,15 +1651,32 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
...
+        size_t vlen = 0;
+        const char *vstr = RedisModule_CallReplyStringPtr(val, &vlen);
+        if (vstr && vlen == 3 && strncmp(vstr, "yes", 3) == 0) {
+            ossClusterRuntime = true;
+        }
+    }
+    if (ceReply) {
+        RedisModule_FreeCallReply(ceReply);
+    }
+    if (rlecVersion && !ossClusterRuntime) {
```
gabsow: Main risk to confirm: this rests on the assumption that a real Redis Enterprise shard never runs with `cluster-enabled yes`. If one ever does, it gets misdetected as OSS, switching onto the internal-secret AUTH path and dropping the `_proxy-filtered` command flags. I believe RLEC keeps `cluster-enabled no` (proxy does routing), but since the whole PR rests on this, worth confirming explicitly. Cheap hardening: log the detected `cluster-enabled` value alongside the existing `Detected redis %s` line.

src/cluster.c
```diff
@@ -415,7 +415,7 @@ static void SendAuthCommandIfNeeded(const struct redisAsyncContext* c, const Nod
         redisAsyncCommand((redisAsyncContext*)c, NULL, NULL, "AUTH %s", n->password);
         return;
     }
-    if (RedisModule_GetInternalSecret && !MR_IsEnterpriseBuild()) {
+    if (RedisModule_GetInternalSecret && clusterCtx.isOss) {
```
gabsow, line 418: After this change there are two enterprise-detection paths that can diverge: `clusterCtx.isOss` (refined, rlec_version + cluster-enabled) here, and `MR_IsEnterprise` (set in mr.c purely from rlec_version) which still reports the private build as enterprise. With these call sites converted, `MR_IsEnterpriseBuild()` now has zero callers (dead code). Suggest either removing it or centralizing the runtime detection so the two flags can't drift. Worth checking no consuming module reads `MR_IsEnterprise` and now disagrees with LibMR's own decision.

src/cluster.c
```diff
@@ -1651,15 +1651,32 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
...
+    /* Note: RedisModule_GetContextFlags() does NOT report
+     * REDISMODULE_CTX_FLAGS_CLUSTER yet at module-load time, so read the
+     * cluster-enabled config directly to learn the runtime mode. */
+    bool ossClusterRuntime = false;
+    RedisModuleCallReply *ceReply = RedisModule_Call(rctx, "config", "cc", "get", "cluster-enabled");
+    if (ceReply && RedisModule_CallReplyType(ceReply) == REDISMODULE_REPLY_ARRAY &&
+        RedisModule_CallReplyLength(ceReply) >= 2) {
+        RedisModuleCallReply *val = RedisModule_CallReplyArrayElement(ceReply, 1);
+        size_t vlen = 0;
+        const char *vstr = RedisModule_CallReplyStringPtr(val, &vlen);
```
gabsow: Minor: the array reply type is checked but the element type isn't before calling `CallReplyStringPtr`. Not a bug (it returns NULL on mismatch and `vstr` is guarded), but line 531 in this file asserts `REDISMODULE_REPLY_STRING` for the same pattern, so this is slightly inconsistent with local style.

gabsow (commented):
Decoupling runtime OSS/enterprise detection from the build/version flag reads correctly for the stated goal: a private build carrying `rlec_version` but running native OSS clustering is now treated as OSS. CONFIG GET handling (RESP2 array, index 1, both replies freed) is correct. A few points inline.

src/cluster.c
AvivDavid23: Done

src/cluster.c
AvivDavid23: Done

src/cluster.c
AvivDavid23: Done

gabsow:
Re-reviewed after `690554b`. All three inline comments are addressed:

- Reply element now guarded with `REDISMODULE_REPLY_STRING` before `CallReplyStringPtr`.
- `MR_IsEnterpriseBuild()`/`MR_IsEnterprise` removed; detection now flows through the single `clusterCtx.isOss` decision (no dangling refs left, build stays clean).
- `Detected redis %s` log now reports `(cluster-enabled=%s)`.

One item the code can't close: this still rests on the assumption that a real RLEC shard never runs with `cluster-enabled yes`. The added log makes a misdetection diagnosable, but worth a verbal confirmation from whoever owns RLEC shard config. Code-level concerns all handled.

AvivDavid23:
> Re-reviewed after `690554b`. All three inline comments are addressed:
> 
> * Reply element now guarded with `REDISMODULE_REPLY_STRING` before `CallReplyStringPtr`.
> * `MR_IsEnterpriseBuild()`/`MR_IsEnterprise` removed; detection now flows through the single `clusterCtx.isOss` decision (no dangling refs left, build stays clean).
> * `Detected redis %s` log now reports `(cluster-enabled=%s)`.
> 
> One item the code can't close: this still rests on the assumption that a real RLEC shard never runs with `cluster-enabled yes`. The added log makes a misdetection diagnosable, but worth a verbal confirmation from whoever owns RLEC shard config. Code-level concerns all handled.

In the fork code, there are evidence that cluster_enabled → "oss-sharding"; etc.. also asm stuff

src/cluster.c
```diff
@@ -1651,15 +1651,36 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
 
     RedisModuleServerInfoData *info = RedisModule_GetServerInfo(rctx, "Server");
     const char *rlecVersion = RedisModule_ServerInfoGetFieldC(info, "rlec_version");
-    if (rlecVersion) {
+    /* Note: RedisModule_GetContextFlags() does NOT report
+     * REDISMODULE_CTX_FLAGS_CLUSTER yet at module-load time, so read the
+     * cluster-enabled config directly to learn the runtime mode. */
+    bool ossClusterRuntime = false;
+    RedisModuleCallReply *ceReply = RedisModule_Call(rctx, "config", "cc", "get", "cluster-enabled");
+    if (ceReply && RedisModule_CallReplyType(ceReply) == REDISMODULE_REPLY_ARRAY &&
```
gabsow: **Reply-shape fragility on RESP3.** `RedisModule_Call` honors the calling context's reply mode. If the module ctx is ever switched to RESP3 (or a future Redis default flips), `CONFIG GET` returns a `REDISMODULE_REPLY_MAP`, not `REDISMODULE_REPLY_ARRAY` — this branch silently fails and `ossClusterRuntime` stays `false`. For an enterprise binary in OSS cluster mode with a RESP3 ctx that means we'd wrongly classify as enterprise and take the (broken) enterprise path — the exact case this PR is fixing.

Two cheap fixes (pick one):
1. Force RESP2 reply parsing by adding `"v"` to the format string: `RedisModule_Call(rctx, "config", "v cc", REDISMODULE_REPLY_ARRAY, "get", "cluster-enabled")` — actually that's not the v flag's meaning. Simpler: drop down to RESP2 by appending an explicit reply-format flag, or
2. Handle both shapes — if `CallReplyType == REDISMODULE_REPLY_MAP`, use `RedisModule_CallReplyMapElement` instead of `CallReplyArrayElement`.

Module ctx defaults to RESP2 today so this is latent, but the failure mode is silent and exactly opposite to the PR's intent.

src/cluster.c
```diff
@@ -1651,15 +1651,36 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
...
+        RedisModule_CallReplyLength(ceReply) >= 2) {
+        RedisModuleCallReply *val = RedisModule_CallReplyArrayElement(ceReply, 1);
+        if (RedisModule_CallReplyType(val) == REDISMODULE_REPLY_STRING) {
+            size_t vlen = 0;
+            const char *vstr = RedisModule_CallReplyStringPtr(val, &vlen);
+            if (vstr && vlen == 3 && strncmp(vstr, "yes", 3) == 0) {
```
gabsow: Minor: `strncmp(vstr, "yes", 3) == 0` with the `vlen == 3` guard is correct (CONFIG GET returns canonical lowercase per Redis spec), but a one-line comment about the case-sensitivity would save the next reader a trip to the Redis source. The analogous code in [RedisTimeSeries PR #80 `detect_bigredis_enabled`](https://github.com/redislabsdev/RedisTimeSeries/pull/80) uses `memcmp` for the same purpose — pick one for cross-repo consistency.

src/cluster.c
```diff
@@ -1651,15 +1651,36 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
...
+                ossClusterRuntime = true;
+            }
+        }
+    }
+    if (ceReply) {
+        RedisModule_FreeCallReply(ceReply);
+    }
+    if (rlecVersion && !ossClusterRuntime) {
```
gabsow: This is the load-bearing line of the PR — the semantic flip from "enterprise iff `rlec_version` is present" to "enterprise iff `rlec_version` is present AND not in OSS cluster mode." Worth a short comment here explaining the case it unblocks (enterprise binary running in OSS cluster mode, e.g. the env0 testing setup) so the rationale isn't only in the PR description. Right now the only documentation that survives review is the `Note: RedisModule_GetContextFlags...` block above, which explains the mechanism but not the *why*.

src/cluster.c
```diff
@@ -1651,15 +1651,36 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
...
         clusterCtx.isOss = false;
     }
     RedisModule_FreeServerInfo(rctx, info);
 
-    RedisModule_Log(rctx, "notice", "Detected redis %s", clusterCtx.isOss? "oss" : "enterprise");
+    RedisModule_Log(rctx, "notice", "Detected redis %s (cluster-enabled=%s)",
```
gabsow, line 1711: The log message `"Detected redis oss (cluster-enabled=yes)"` is parseable but the `cluster-enabled` axis adds noise to operators who don't care about the OSS-cluster-on-enterprise case. Two options:
1. Always log it — current behavior, slight noise on the 99% normal case.
2. Log the cluster-enabled state only when it disagrees with the rlec_version signal (i.e. "enterprise binary in OSS cluster mode" or "OSS binary that somehow has rlec_version" — really just the first).

No wrong answer; option 2 is more idiomatic for "interesting cases only" logging.

src/cluster.c
```diff
@@ -1651,15 +1651,36 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
 
     RedisModuleServerInfoData *info = RedisModule_GetServerInfo(rctx, "Server");
```
gabsow: Cosmetic: `MR_GetRedisVersion()` (src/mr.c:1546-ish, called before `MR_ClusterInit`) already parses `rlec_version` from the same INFO Server reply for `MR_RlecMajorVersion` etc. After this PR, we're parsing it again here. Worth a follow-up to surface `rlec_version` presence into a single field (or just `MR_RlecMajorVersion >= 0`) and drop the second `GetServerInfo` call. Not blocking.

gabsow (commented):
## Review of LibMR PR #98 (MOD-15896)

Focused 118-line change with a clear single purpose: stop using the load-time `MR_IsEnterprise` global, derive OSS/enterprise classification at `MR_ClusterInit` from the **runtime combination of `rlec_version` AND `cluster-enabled`**. The motivating use case (also surfaces in [RedisTimeSeries PR #80](https://github.com/redislabsdev/RedisTimeSeries/pull/80)): an Enterprise-build redis-server running in OSS cluster mode (no RLEC proxy) now takes the OSS code path instead of the broken enterprise-with-no-proxy path.

### Behavior summary

| Build | `cluster-enabled` | Old `isOss`? | New `isOss`? |
|---|---|---|---|
| OSS redis | n/a (no rlec_version) | true | true ✓ unchanged |
| Enterprise redis, RLEC-managed | no | false | false ✓ unchanged |
| Enterprise redis, OSS-cluster mode | **yes** | false (broken) | **true** ← this PR's fix |

The four downstream knobs that flip on `clusterCtx.isOss` (internal-secret AUTH, HELLO retry, command ACL flag `_proxy-filtered` vs `internal`, REFRESHCLUSTER registration) all make sense for the new OSS-cluster-on-enterprise-binary case.

### Real concerns — inline

One medium (RESP3 reply shape), the rest low / nits. Nothing blocking.

### Approve / comment / request-changes

**Comment.** The design is sound, the diff is contained, the fix unblocks the test infra. Inline comments are fixable in a follow-up commit.

src/cluster.c
AvivDavid23: Done in `ecd4eb1`. Now handles both shapes: RESP2 `REDISMODULE_REPLY_ARRAY` (element 1) and RESP3 `REDISMODULE_REPLY_MAP` (via `RedisModule_CallReplyMapElement`, guarded for the symbol's presence). So the silent-fail-to-enterprise path can't trigger if the module ctx ever flips to RESP3.

src/cluster.c
AvivDavid23: Done in `ecd4eb1`. Switched to `memcmp` to match `detect_bigredis_enabled` in RTS PR #80, and added a comment noting CONFIG GET returns canonical lowercase so a case-sensitive 3-byte compare is safe.

src/cluster.c
AvivDavid23: Done in `ecd4eb1`. Added a comment explaining this is the load-bearing decision and the case it unblocks (enterprise binary running as an OSS cluster, e.g. the env0 testing setup → must take the OSS path).

src/cluster.c
AvivDavid23: Kept option 1 (always log). This was the hardening we agreed to add last round, and an always-present `cluster-enabled=` axis keeps the log line greppable/uniform for tooling. The noise is one token on the normal case; happy to switch to conditional if you feel strongly.

src/cluster.c
AvivDavid23: Done in `ecd4eb1` (folded into this PR rather than a follow-up). Dropped the second `GetServerInfo("Server")` and reuse `MR_RlecMajorVersion >= 0`, which `MR_GetRedisVersion()` already populates before `MR_ClusterInit` runs. Removes the duplicate parse and the divergence risk.

src/cluster.c
```diff
@@ -1649,17 +1649,55 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
...
+     * parsed into MR_RlecMajorVersion by MR_GetRedisVersion() (run before
+     * MR_ClusterInit), so reuse it here instead of fetching Server info a
+     * second time; MR_RlecMajorVersion >= 0 means the field was present.
+     *
+     * Treat the shard as enterprise only when rlec_version is present AND we
+     * are not in OSS cluster mode. This is the load-bearing decision of the
+     * PR: it unblocks an enterprise binary running as an OSS cluster (e.g. the
+     * env0 testing setup), which must take the OSS path (internal-secret AUTH,
+     * no _proxy-filtered command flags) rather than the enterprise path. */
+    if (MR_RlecMajorVersion >= 0 && !ossClusterRuntime) {
```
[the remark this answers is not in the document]
AvivDavid23: Done in `5ebd3cf`. Enterprise detection now keys on **presence** of `rlec_version` (new `MR_RlecVersionPresent`, set from the `strstr` in `MR_GetRedisVersion` independent of the `sscanf`) rather than `MR_RlecMajorVersion >= 0`. A present-but-unparseable `rlec_version` no longer reads as OSS — this restores the pre-PR presence semantics while keeping the single Server-info parse (no second `GetServerInfo`).

src/cluster.c
```diff
@@ -1649,17 +1649,55 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
...
+        if (val && RedisModule_CallReplyType(val) == REDISMODULE_REPLY_STRING) {
+            size_t vlen = 0;
+            const char *vstr = RedisModule_CallReplyStringPtr(val, &vlen);
+            /* CONFIG GET returns the canonical lowercase "yes"/"no" (Redis
+             * spec), so a case-sensitive compare of exactly 3 bytes is safe. */
+            if (vstr && vlen == 3 && memcmp(vstr, "yes", 3) == 0) {
+                ossClusterRuntime = true;
+            }
+        }
+        RedisModule_FreeCallReply(ceReply);
```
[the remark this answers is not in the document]
AvivDavid23: Done in `5ebd3cf`. The parse now tracks whether `CONFIG GET` resolved to an actual string value (`ceResolved`); on an error reply, unexpected shape, or empty value we log a `warning` instead of failing silently. Kept the conservative default (`ossClusterRuntime` stays false → an `rlec_version` shard is still treated as enterprise as before): a transient `CONFIG GET` failure shouldn't drop a real RLEC shard onto the OSS internal-secret AUTH path. The misclassification is now diagnosable from the log rather than silent.

gabsow (approved):
All 5 inline findings addressed in `ecd4eb1d` + `5ebd3cf3`, verified against the current head:

| Finding | Fix |
|---|---|
| RESP3 reply-shape fragility | Both shapes now handled — `REPLY_ARRAY` (RESP2) via `CallReplyArrayElement`, `REPLY_MAP` (RESP3) via `CallReplyMapElement` with symbol-presence guard (`src/cluster.c:1665-1669`). Silent-fail-to-enterprise path can't trigger. |
| `strncmp` vs `memcmp` cross-repo consistency | Switched to `memcmp` to match `detect_bigredis_enabled` in RTS PR #80, with the canonical-lowercase comment (`src/cluster.c:1676-1678`). |
| Load-bearing comment on the semantic flip | Multi-line block at `src/cluster.c:1695-1706` explains the env0 case and why it's the load-bearing decision. |
| Always-log `cluster-enabled` | Reasoned defense accepted — uniform grep-able log line is worth the one extra token on the normal case. |
| Duplicate `GetServerInfo("Server")` parse | Removed. Now reuses `MR_RlecVersionPresent` (new flag in `src/mr.c:1552`, set from `strstr` presence independently of `sscanf`), so `MR_ClusterInit` no longer re-fetches Server info. The presence-vs-parsed-version distinction (`5ebd3cf`) also correctly handles a present-but-unparseable `rlec_version`. |

Bonus: the `ceResolved` + warning log on transient `CONFIG GET` failure (`src/cluster.c:1685-1694`) is a nice hardening — conservative default keeps the previous classification on failure but the misclassification is now diagnosable from the log rather than silent.

Call-order assumption (`MR_GetRedisVersion()` runs before `MR_ClusterInit`) verified at `src/mr.c:1570-1572`.

LGTM. :ship:

PR #99, opened 2026-06-15
MOD-16549 Return an error from short-form CLUSTERSET when the slot-ranges API is missing

gabsow:
## Problem

The short form of `CLUSTERSET` relies on `RedisModule_GetClusterNodeSlotRanges` ([redis/redis#14953](https://github.com/redis/redis/pull/14953)). When the host Redis build does not export it, [#94](https://github.com/RedisGears/LibMR/pull/94) made `SetClusterDataShortForm` log a warning and return early to avoid a crash — **but the command still replied `+OK`** while leaving the cluster unconfigured.

This breaks the planned DMC rollout strategy: DMC wants to send the short form and **fall back to the long form on error**. With a silent `+OK`, DMC can't tell "configured" from "module silently no-op'd on an old core," so it never falls back and the cluster stays unconfigured.

For comparison, RediSearch already returns an error in this case (its short-form guard is `argc == 3 && RedisModule_GetClusterNodeSlotRanges`, which short-circuits to the long-form parser → `ERROR_MISSING("MYID")` → error reply).

## Fix

Propagate a status out of `SetClusterDataShortForm` → `MR_SetClusterData` → `ClusterSetCtx`, and have `MR_ClusterSetUnblock` reply with an error (`-ERRCLUSTER Failed to set cluster topology`) instead of `+OK` when configuration failed. Follows the existing `MessageCtx.reply` deferred-reply idiom used by the inner-communication path.

Covers both failure paths in the short form: the missing-API case and the `RedisModule_GetClusterNodesList` failure. The success path (API present) is unchanged — still replies `+OK`. Applies to both `TIMESERIES.CLUSTERSET` and `…CLUSTERSETFROMSHARD`, which share `MR_ClusterSetInternal`.

## Test

Adds `testShortFormClusterSetWithoutSlotRangesApi` to `test_errors.py`, gated to Redis < 8.0 (the `7.2`/`7.4` CI legs, which predate the API), asserting the short form returns an error. The success path is covered end-to-end by RedisTimeSeries' `test_short_form_clusterset` flow test (runs in `oss-cluster` against a build that exports the API).

Built locally (`make libmr`); `cluster.c` compiles clean.

---

> [!NOTE]
> **Medium Risk**
> Touches cluster bootstrap for CLUSTERSET/CLUSTERSETFROMSHARD; behavior change only on previously failing short-form paths, but mis-detection could affect rollout automation.
> 
> **Overview**
> **Short-form `CLUSTERSET` no longer replies `OK` when topology cannot be applied.** `SetClusterDataShortForm` and `MR_SetClusterData` now return `REDISMODULE_OK` / `REDISMODULE_ERR` instead of void, covering a missing `RedisModule_GetClusterNodeSlotRanges` and failure to obtain the cluster nodes list.
> 
> Failures are stored on `ClusterSetCtx.errReply` and surfaced in `MR_ClusterSetUnblock` as `-ERRCLUSTER Failed to set cluster topology`, so callers (e.g. DMC) can retry with long-form `CLUSTERSET`. Successful short-form behavior is unchanged.
> 
> A pytest on Redis &lt; 8.0 asserts the error when the slot-ranges API is absent, plus existing arity validation for malformed short-form args.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3179741e15be243a5ed02b0823b3035d9f076af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -172,6 +172,7 @@ typedef struct ClusterSetCtx {
     RedisModuleString **argv;
     int argc;
     bool force;
+    const char* errReply;  // NULL => reply "OK"; otherwise reply this error to the client
 }ClusterSetCtx;
 
 typedef enum MessageReply {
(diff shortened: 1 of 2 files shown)


src/cluster.c
```diff
@@ -172,6 +172,7 @@ typedef struct ClusterSetCtx {
     RedisModuleString **argv;
     int argc;
     bool force;
+    const char* errReply;  // NULL => reply "OK"; otherwise reply this error to the client
```
galcohen-redislabs, line 175: See my comment [here](https://redis.slack.com/archives/C09EK7JR61Y/p1781528873613649?thread_ts=1779785382.821159&cid=C09EK7JR61Y)

gabsow: I decided to go with error insted of a crash , I can push for actually returning an error as reply to the claster set message so it will be easier to be catched but not to crash , so it will not somehow crash in the field.

galcohen-redislabs: Problem is: with this change alone (i.e., without adding code to handle it in the dmc, etc) it could lead to data corruption.

galcohen-redislabs (approved):

galcohen-redislabs (approved):

PR #100, opened 2026-06-18
MOD-16382 Add MR_ClusterRefreshTopology for event-driven OSS topology refresh

gabsow:
## Summary

Adds **`MR_ClusterRefreshTopology()`** so a consumer module can refresh the LibMR cluster view in response to a Redis server event, instead of requiring an operator to call a manual refresh command (e.g. `TIMESERIES.REFRESHCLUSTER`) on every primary after each reshard.

### What it does

- `MR_ClusterRefreshTopology()` schedules the existing `MR_RefreshClusterData()` on the LibMR event loop (the thread-safe place to mutate topology), with **no blocked client** to reply to — the difference from the command-driven `MR_ClusterRefresh`.
- It is **gated by `clusterCtx.isOss`**, so it is a no-op outside OSS cluster mode (under Enterprise the DMC drives topology via `CLUSTERSET`), mirroring the OSS-only registration of `REFRESHCLUSTER`. Safe to call unconditionally from an event callback.
- Declared in `cluster.h` next to the other `MR_Cluster*` entry points.

The local `redismodule.h` copy is updated with the `RedisModuleEvent_ClusterTopologyChange` definitions (id 20) so LibMR consumers compile against the new event; this mirrors redis/redis#15350 and will be reconciled by the normal header sync.

Used by the RedisTimeSeries consumer PR (auto-refresh on cluster topology change). Verified end-to-end on a 3-node OSS cluster: topology auto-refreshes on formation and reshard with no manual command.

Depends on / mirrors: redis/redis#15350
Relates-to: MOD-9152, RED-148990

> Draft — gated on the core event being finalized.

---

> [!NOTE]
> **Medium Risk**
> Changes OSS cluster slot routing and when inter-shard connections are torn down; incorrect reconcile could misroute cross-shard traffic until a full refresh runs.
> 
> **Overview**
> Adds **`MR_ClusterRefreshTopology()`** so consumer modules can refresh LibMR’s cluster view from **`RedisModuleEvent_ClusterTopologyChange`** (and related server events) without a manual **`REFRESHCLUSTER`** on every primary. The call is **OSS-only** (`clusterCtx.isOss`), ignores advisory `change_flags`, and queues work on the LibMR event loop so it is safe from a Redis server-event callback.
> 
> Reconciliation is **`MR_UpdateClusterSlots()`**: it builds a master/slot view from the same cluster module APIs as short-form **`CLUSTERSET`** (`BuildClusterApiView`). When the master set and addresses are unchanged, **`MR_TryApplyTopologyInPlace()`** clears and repoints **`slots[]`** while keeping existing shard connections; otherwise it falls back to the existing full **`MR_RefreshClusterData()`** teardown/rebuild.
> 
> The vendored **`redismodule.h`** gains **`RedisModuleEvent_ClusterTopologyChange`**, topology change reason flags, and **`RedisModuleClusterTopologyChangeInfo`** so LibMR consumers can compile against the new event (aligned with redis/redis#15350).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f4bc06e189ddecaf64cbf736e536cd192f713ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -957,7 +957,7 @@ static void MR_RefreshClusterData(){
 
 /* Reconcile the LibMR cluster view with a fresh CLUSTER SLOTS, reusing the existing
  * Node structs (and their live connections) wherever possible. This is the single
- * entry point the debounced topology-change handler calls, regardless of which
+ * entry point the topology-change handler calls, regardless of which
  * reason flags the event carried, and it decides what to do from the actual
  * primary-set delta:
  *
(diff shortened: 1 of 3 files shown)


gabsow:
Verified end-to-end against redis/redis#15350 on a live 4-shard scale-out (3-shard OSS cluster + RedisTimeSeries, then add a 4th shard and reshard 4096 slots with key migration):

| node | raw topology events | actual refreshes |
|---|---|---|
| 7811 (src) | 4097 | **2** |
| 7812 | 8193 | **2** |
| 7813 | 8193 | **2** |
| 7814 (new) | 4093 | **1** |

Cross-shard `TS.QUERYINDEX`/`TS.MRANGE` returned the complete result (24/24, including series migrated to the brand-new shard) on every node, with **no `TIMESERIES.REFRESHCLUSTER`**. The trailing debounce collapses the per-slot `CLUSTER SETSLOT` event storm (\~8k events) into 1-2 refreshes, avoiding thousands of connection teardowns.

gabsow:
@galcohen-redislabs — design decision for you: do we keep the topology-refresh debounce, and in what shape?

**What it is:** every topology event just bumps a counter; the actual reconcile (`MR_UpdateClusterSlots`: GIL lock + `CLUSTER SLOTS` + slot-map repoint) runs only after 100 ms pass with no new events (`MR_TOPOLOGY_DEBOUNCE_MS`, trailing-edge).

**Why it exists:** to collapse event bursts into one reconcile. Historically it also prevented N connection teardowns — but the primary-set diff in this PR now handles that (rebuild only when a primary enters/leaves), so today the debounce only prevents reconcile churn, not command drops.

**Measured event volume per action** (live 3x3, patched core):

| action | events per node | debounce benefit |
|---|---|---|
| ASM range migration (101 slots) | exactly 1, at takeover; 0 on abort | none — pure +100 ms convergence delay |
| failover / fail-back | 1–2 | ~none |
| add replica (MEET + REPLICATE) | 2–3, seconds apart | ~none (separate quiet windows) |
| legacy per-slot SETSLOT reshard | ~1 per slot per master | real: N reconciles → 1 |

**The cost of keeping it as-is:** after an ASM takeover we keep routing on the stale map for up to 100 ms — cross-shard commands hitting moved slots get the transient unavailable-slots error until the reconcile runs. Correctness holds (verified in test_asm.py — only whitelisted transient errors, results converge), but the window is avoidable.

**Options:**
1. **Leading + trailing (my recommendation):** reconcile immediately on the first event of a burst, plus one trailing reconcile after 100 ms of quiet. ASM/failover converge with zero added latency; a legacy burst costs ~2 reconciles instead of N. ~10-line change.
2. **Trailing-only (current):** simplest, but pays 100 ms on every ASM migration for a burst that, per the measurements, doesn't exist there.
3. **No debounce:** fastest convergence everywhere; a legacy 1,000-slot rebalance would run ~1,000 GIL-locked `CLUSTER SLOTS` reconciles over the reshard window.

Your call — happy to implement whichever you pick.

gabsow:
Per the review call (keep it simple): **b35bab1** drops the debounce entirely — every event now schedules one reconcile directly on the event loop; the `cluster.h` comment is updated to match the primary-set-diff behavior.

Self-review also caught a gap in the in-place path: a primary restarting at a **new address with the same node id** (pod reschedule keeping nodes.conf) kept its stale ip/port, so the reconnect loop would redial the dead address forever. **d3830c1** compares the address from the fresh `CLUSTER SLOTS` (port via `RedisModule_GetClusterNodeInfo`, same as `MR_RefreshClusterData`) and escalates to a full rebuild on mismatch.

`test_asm` 5/5 green on OSS cluster with the patched core after both commits.

src/cluster.c
```diff
@@ -955,6 +955,138 @@ static void MR_RefreshClusterData(){
...
+ * Deciding by the primary-set delta rather than by the event's reason flags means an
+ * over-broad or spurious notification (a replica re-pointing, an OK<->FAIL flip, a
+ * slotless node joining) that does not change the primary set costs only a cheap
+ * in-place repoint -- never a connection rebuild or a dropped in-flight multi-key
+ * command.
+ *
+ * In OSS mode CurrCluster->slots is the routing source of truth (the SendMsgType_BySlot
+ * path) and fan-out (SendMsgType_All) iterates the nodes dict; both stay correct here.
+ * Per-node slotRanges are only consumed by the Enterprise CLUSTERSET build and the
+ * debug RG.INFOCLUSTER reply, so they are intentionally left untouched. */
```
galcohen-redislabs: I don't see any point in this long and mostly confusing comment (what's an in-place reshard? why mention oss-mode? or in-flight multi-key commands?)

src/cluster.c
```diff
@@ -955,6 +955,138 @@ static void MR_RefreshClusterData(){
...
+        MR_RefreshClusterData();
+        return;
+    }
+
+    if(!(RedisModule_GetContextFlags(mr_staticCtx) & REDISMODULE_CTX_FLAGS_CLUSTER)){
+        return;
+    }
+
+    RedisModule_ThreadSafeContextLock(mr_staticCtx);
+    RedisModuleCallReply *allSlotsReply = RedisModule_Call(mr_staticCtx, "cluster", "c", "slots");
```
galcohen-redislabs: We already have the `RedisModule_GetClusterNodeSlotRanges()`. Why call (and parse) redis commands?

src/cluster.c
```diff
@@ -955,6 +955,138 @@ static void MR_RefreshClusterData(){
...
+ * over-broad or spurious notification (a replica re-pointing, an OK<->FAIL flip, a
+ * slotless node joining) that does not change the primary set costs only a cheap
+ * in-place repoint -- never a connection rebuild or a dropped in-flight multi-key
+ * command.
+ *
+ * In OSS mode CurrCluster->slots is the routing source of truth (the SendMsgType_BySlot
+ * path) and fan-out (SendMsgType_All) iterates the nodes dict; both stay correct here.
+ * Per-node slotRanges are only consumed by the Enterprise CLUSTERSET build and the
+ * debug RG.INFOCLUSTER reply, so they are intentionally left untouched. */
+static void MR_UpdateClusterSlots(){
```
galcohen-redislabs: In the context of LibMR we only care about the set of live master nodes (i.e., since we simply call all of them and don't care for slots at all). Why do anything more complex than that?
Also - short-form CLUSTERSET already builds when we need, so better to refactor it (e.g., to return a data struct that represents the set of primary nodes and is comparable with other objects of that struct) than adding a new function.
(we might want a refined logic of "all master nodes _except slotless ones_" but honestly I think this is an unnecessary optimization; especially since slotless nodes are added to a cluster and immediately get some slots migrated towards them).

src/cluster.c
gabsow: Trimmed in 6118f92.

src/cluster.c
gabsow: Right — switched to `GetClusterNodesList`/`GetClusterNodeInfo`/`GetClusterNodeSlotRanges` (the short form's source) in 6118f92. Also drops the CLUSTER SLOTS non-TLS-port workaround. Falls back to a full refresh when the API is unavailable.

src/cluster.c
gabsow: Done in 6118f92: the view is now a comparable struct (`TopologyViewEntry`) built by `BuildClusterApiView()` from the same source the short form uses, and the decision is a plain master-set compare — ids + addresses (an address change must reconnect, so it escalates). Slotless masters are skipped per your note. The slot-map repoint stays as the cheap apply step since LibMR's public API exposes slot routing (`SendMsgBySlot`/`IsMySlot`). #104 is being restacked to move both CLUSTERSET forms onto this same mechanism, as you suggested.

PR #104, opened 2026-07-06
MOD-16399 Skip the rebuild when CLUSTERSET carries an unchanged topology

gabsow:
## Summary

- Handle only long-form `CLUSTERSET`.
- Before parsing or building cluster structures, compare the incoming arguments with the command stored on the current cluster.
- Ignore the command name and `MYID`; every topology-bearing argument must match.
- Preserve the current cluster, connections, and run ID for an identical command. Any changed argument follows the existing full rebuild path.
- Leave short-form `CLUSTERSET`, Redis core topology events, and the existing internal cluster comparison unchanged from `master`.

The final diff changes two files: 25 production lines plus focused test coverage. The previously published commit history is preserved.

## Tests

- Identical long-form no-op test: passed.
- `testMassiveClusterSet`: 1,000 alternating changed topologies passed.
- Full standalone `test_network.py`: 19 tests completed, 0 failed.
- RedisTimeSeries integration coverage: [RedisTimeSeries#2088](https://github.com/RedisTimeSeries/RedisTimeSeries/pull/2088).

---

> [!NOTE]
> **Medium Risk**
> Touches cluster topology application in libMR; incorrect equality could skip needed rebuilds or miss updates, though comparison is narrow and tests cover identical vs changed args.
> 
> **Overview**
> Long-form **CLUSTERSET** no longer tears down and rebuilds cluster state when the incoming topology matches what is already stored. Before parsing, **`IsSameLongFormClusterSet`** compares every argument except **MYID** (index 6) against the saved **`clusterSetCommand`**, using **`RedisModule_StringPtrLen`** lengths so embedded NUL bytes are compared correctly. A match logs and returns **OK** without changing connections or the cluster **run ID**; any differing argument still goes through **`SetClusterDataLongForm`**. Short-form **CLUSTERSET** and other topology paths are unchanged.
> 
> Tests add **`testIdenticalLongFormClusterSetIsNoOp`** (stable run ID on repeat, rebuild on byte-level arg change and shard-id change) and adjust **`testMassiveClusterSet`** to alternate mock shard IDs so repeated updates still exercise rebuilds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d1759e6e853d5f248173fa39dd74d800fc4d41a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -957,7 +957,7 @@ static void MR_RefreshClusterData(){
 
 /* Reconcile the LibMR cluster view with a fresh CLUSTER SLOTS, reusing the existing
  * Node structs (and their live connections) wherever possible. This is the single
- * entry point the debounced topology-change handler calls, regardless of which
+ * entry point the topology-change handler calls, regardless of which
  * reason flags the event carried, and it decides what to do from the actual
  * primary-set delta:
  *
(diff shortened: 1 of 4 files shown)


src/cluster.c
```diff
@@ -1252,7 +1252,35 @@ static void SetClusterDataLongForm(RedisModuleString** argv, int argc){
...
+static bool ClusterSetCommandIsUnchanged(RedisModuleString** argv, int argc){
+    Cluster* cur = clusterCtx.CurrCluster;
+    if (!cur || !cur->clusterSetCommand || cur->clusterSetCommandSize != argc)
+        return false;
+    if (!IsLongFormClusterSet(argc))
+        return false;
+    for (int i = 1 ; i < argc ; ++i) {
+        if (i == CLUSTERSET_MYID_LONG_FORM_INDEX)
+            continue;
+        const char* arg = RedisModule_StringPtrLen(argv[i], NULL);
```
galcohen-redislabs: To be on the safe side I would add the length (i.e., to cover cases of non null-terminated strings) and also use it in the comparison below.

src/cluster.c
```diff
@@ -1252,7 +1252,35 @@ static void SetClusterDataLongForm(RedisModuleString** argv, int argc){
...
+        if (strcmp(arg, cur->clusterSetCommand[i]) != 0)
+            return false;
```
galcohen-redislabs, line 1424: I don't like the double negative (i.e., the function is checking *un*changing and returns false if changed). Better to have a positive function name (e.g., `ClustersetIsNewTopology()`) and adjust the logic.

src/cluster.c
```diff
@@ -1252,7 +1252,35 @@ static void SetClusterDataLongForm(RedisModuleString** argv, int argc){
     mr_dictEmpty(clusterCtx.nodesMsgIds, NULL);
 }
 
+/* Returns true when the incoming long-form cluster-set command carries the exact
+ * topology the current cluster was built from. The MYID slot is excluded: it names
```
galcohen-redislabs: Well.. for the same reasons we should do the same for the short-form (which probably means we need to refactor `SetClusterDataShortForm()` such that the topology data (or at least the part that's relevant for us) is returned and could be released and be compared to the one generated by the next short-form clusterset).
Fortunately, we can skip the case of comparing short to long form topology results, since no version of a module should get both formats.

src/cluster.c
gabsow: Done in 6f7b604 — the arguments are now compared via `RedisModule_StringPtrLen` length + `memcmp` (a stored copy that was truncated at an embedded NUL also compares unequal, which errs on the rebuild side).

src/cluster.c
gabsow: Done in 6f7b604 — renamed to `ClusterSetIsNewTopology` and flipped the logic (the skip is now `if (!ClusterSetIsNewTopology(...))`).

src/cluster.c
gabsow: Done in 6f7b604. Rather than returning the intermediate node structures from `SetClusterDataShortForm`, the derived master set is serialized into a canonical snapshot string (one `<id> <ip>:<port> <slot-ranges>` line per master, sorted by node id so the nodes-list iteration order doesn't matter), stored on the `Cluster`, and compared against the snapshot freshly derived for the next short-form command. The AUTH password is intentionally not part of the snapshot — it arrives in the arguments, which are compared separately (so a password change still rebuilds). Long-vs-short comparison is naturally excluded by the argc check, per your note.

Verified empirically on a live 2-node OSS cluster: initial short form rebuilds, identical re-send skips, re-send after a `CLUSTER SETSLOT` move rebuilds (fresh snapshot differs from the stored one), identical re-send skips again. Full pytest suite still passes (34 run, 0 failed).

gabsow:
**Restacked** per the review direction on #100 (and today's discussion): this PR now sits **on top of #100** and moves both CLUSTERSET forms onto the **same reconcile mechanism** the topology-change event uses, instead of a separate string-snapshot compare.

What changed vs. the previous head:
- The short-form `ShortFormTopologySnapshot` string compare is **gone** — the short form now reduces the server's cluster state to the shared comparable view (`TopologyViewEntry` via `BuildClusterApiView`, from #100) and reconciles semantically.
- The long form keeps the identical-args fast path, and any *changed* args are first reconciled through the same mechanism: **a reshard that only moves slot ranges between the same shards is now applied in place** — connections, in-flight executions and the run id survive (previously any non-identical payload rebuilt). A shard entering/leaving the set, or changing its address or credentials, still tears down and rebuilds.
- Tests updated to the new spec + new coverage: moved-boundary → in-place (run id + live connection survive); re-send of moved args → no-op; changed shard id / changed password → rebuild; `testMassiveClusterSet` now alternates the shard id so it still stresses 1000 genuine rebuilds.

Validated: RTS `test_asm` 5/5 (incl. short-form path) and the MOD-16399 flow test (identical → run id kept; changed address → rebuilt) against this head.

Note: until #100 merges, this PR's diff-vs-master shows #100's commits too — review just the top commits (`57a2dd1`, `e8a3578`, `fafda41`).

gabsow:
CI hang root-caused (docker repro): the rewritten `testMassiveClusterSet` alternation started at shard id '2' — argument-identical to the topology `__enter__` had just applied — so the identical-args skip made iteration 0 a no-op and `GetConnection()` blocked forever on its timeout-less queue (the 45-min silent cancel). Fixed in 1b5d6c3: alternation starts at '3', and `GetConnection` now has a finite timeout so any such regression fails fast instead of wedging the job. The skip behaved correctly — the test's phase was wrong.

src/mr.c
```diff
@@ -1526,6 +1559,29 @@ void MR_FreeExecution(Execution* e) {
...
+    //    integrity (MR_BuildCluster() returns non-NULL only if the topology is valid).
+    // 2. The event-loop thread (in the lock-protected MR_UpdateClusterTopologyIfNeeded): compare the
+    //    candidate against the installed clusterCtx.CurrCluster and swap it in
+    //    (alongside other fields, e.g., .clusterSize) only if it changed.
+    //    Since this step reads and writes the cluster context it must run on the event-loop thread.
+    //
+    // Because the comparison runs later on the event-loop thread, this function
+    // cannot return a boolean indicating whether the topology actually changed or not
+    // so it returns nothing.
+    Cluster *cluster = MR_BuildCluster(argv, argc, password);
```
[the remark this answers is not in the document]
gabsow: Thanks. This finding is against the superseded commit `20bc188`. The current PR head (`c23cc6a`) no longer changes `MR_UpdateClusterTopology` or the topology-building path; the referenced code is already present on `master`.

The concern may be valid independently for a Redis build that enables topology events without exporting `RedisModule_GetClusterNodeSlotRanges`, but it belongs to the topology-event mechanism rather than this long-form `CLUSTERSET` no-op change. I am keeping #104 limited to the long-form argc/argv comparison and its tests, so no code change is needed here.

src/cluster.c
```diff
@@ -1406,8 +1406,31 @@ static void SetClusterDataLongForm(RedisModuleString** argv, int argc){
     mr_dictEmpty(clusterCtx.nodesMsgIds, NULL);
 }
 
+/* MYID identifies the receiving shard and is not retained in clusterSetCommand. */
```
galcohen-redislabs: Would be better to put this comment near the relevant `if ... continue;` statement below.

gabsow: Done in `4d1759e` — moved the MYID explanation directly above the relevant `if ... continue;` guard.

galcohen-redislabs (approved):

PR #105, opened 2026-07-14
MOD-16382 Handle topology-change notifications

galcohen-redislabs:
First step - refactor SetClusterDataShortForm(): extract code into a general BuildCluster() function and call it.
Note that this required some additional cleanups (i.e., don't rely too much on clusterCtx)

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -785,33 +785,37 @@ static void MR_NodeFree(Node* n){
     MR_NodeFreeInternals(n);
 }
 
-static void MR_ClusterFree(){
-    MR_AbortRunningExecutions();
-
-    if(clusterCtx.CurrCluster->myId){
-        MR_FREE(clusterCtx.CurrCluster->myId);
+static void FreeCluster(Cluster* cluster){
+    if(cluster->myId){
+        MR_FREE(cluster->myId);
     }
-    if(clusterCtx.CurrCluster->nodes){
-        mr_dictIterator *iter = mr_dictGetIterator(clusterCtx.CurrCluster->nodes);
+    if(cluster->nodes){
+        mr_dictIterator *iter = mr_dictGetIterator(cluster->nodes);
         mr_dictEntry *entry = NULL;
         while((entry = mr_dictNext(iter))){
             Node* n = mr_dictGetVal(entry);
             MR_NodeFree(n);
         }
         mr_dictReleaseIterator(iter);
-        mr_dictRelease(clusterCtx.CurrCluster->nodes);
+        mr_dictRelease(cluster->nodes);
     }
 
-    if(clusterCtx.CurrCluster->clusterSetCommand){
-        for(int i = 0 ; i < clusterCtx.CurrCluster->clusterSetCommandSize ; ++i){
-            if(clusterCtx.CurrCluster->clusterSetCommand[i]){
-                MR_FREE(clusterCtx.CurrCluster->clusterSetCommand[i]);
+    if(cluster->clusterSetCommand){
+        for(int i = 0 ; i < cluster->clusterSetCommandSize ; ++i){
+            if(cluster->clusterSetCommand[i]){
+                MR_FREE(cluster->clusterSetCommand[i]);
             }
         }
-        MR_FREE(clusterCtx.CurrCluster->clusterSetCommand);
+        MR_FREE(cluster->clusterSetCommand);
     }
 
-    MR_FREE(clusterCtx.CurrCluster);
+    MR_FREE(cluster);
+}
+
+static void MR_ClusterFree(){
+    MR_AbortRunningExecutions();
+
+    FreeCluster(clusterCtx.CurrCluster);
     clusterCtx.CurrCluster = NULL;
     clusterCtx.minSlot = 0;
     clusterCtx.maxSlot = 0;
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -145,7 +145,7 @@ typedef struct Node{
 
 typedef struct Cluster {
     char* myId;
-    mr_dict* nodes;  // Note: we only keep master nodes here
+    mr_dict* nodes;  // Note: we only keep master nodes here (including slotless master nodes, if any)
     Node* slots[NUMBER_OF_SLOTS];
     size_t clusterSetCommandSize;
     char** clusterSetCommand;
@@ -1141,11 +1141,24 @@ Cluster* BuildCluster(RedisModuleString** argv, int argc, const char* password) 
             maxSlot = slots->ranges[j].end;
             if (j > 0)  // The 0 case is handled by the MR_CreateNode() above
                 mr_listAddNodeTail(aMasterNode->slotRanges, NewSlotRange(minSlot, maxSlot));
-            for (int k = minSlot ; k <= maxSlot ; k++)
+            for (int k = minSlot ; k <= maxSlot ; k++) {
+                if (cluster->slots[k] != NULL) {
+                    RedisModule_Log(mr_staticCtx, "warning",
+                        "Slot %d is claimed by two master nodes: %s and %s; rejecting topology",
+                        k, cluster->slots[k]->id, aMasterNode->id);
+                    FreeCluster(cluster);
+                    cluster = NULL;
+                    break;
+                }
                 cluster->slots[k] = aMasterNode;
+            }
+            if (cluster == NULL)
+                break;
         }
 
         RedisModule_ClusterFreeSlotRanges(mr_staticCtx, slots);
+        if (cluster == NULL)
+            break;
     }
     RedisModule_FreeClusterNodesList(nodeList);
 
@@ -1157,6 +1170,7 @@ Cluster* BuildCluster(RedisModuleString** argv, int argc, const char* password) 
 // there, or scheduled as a task from MR_UpdateClusterTopology on the main thread).
 void MR_UpdateClusterTopologyInternal(void* ctx){
     Cluster* cluster = ctx;
+    RedisModule_Assert(cluster != NULL);
 
     clusterCtx.CurrCluster = cluster;
     memcpy(clusterCtx.myId, cluster->myId, REDISMODULE_NODE_ID_LEN + 1);
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -938,7 +938,7 @@ static void MR_RefreshClusterData(){
         Node* n = MR_GetNode(clusterCtx.CurrCluster, nodeId);
         if(!n){
             /* If we have internal secret we will ignore the clusterCtx.password, we do not need it. */
-            n = MR_CreateNode(clusterCtx.CurrCluster, nodeId, nodeIp, (unsigned short)port, RedisModule_GetInternalSecret ? NULL : clusterCtx.password, NULL, minSlot, maxSlot);
+            n = MR_CreateNode(clusterCtx.CurrCluster, nodeId, nodeIp, (unsigned short)port, MR_ClusterGetPassword(), NULL, minSlot, maxSlot);
         }
 
         if (n->isMe) {
@@ -1168,7 +1168,7 @@ static int SetClusterDataShortForm(RedisModuleString** argv, int argc){
     const char *password = NULL;
     switch (argc) {
     case 1:
-        password = NULL;
+        password = MR_ClusterGetPassword();
         break;
     case 3: {
         const char *token = RedisModule_StringPtrLen(argv[1], NULL);
@@ -1835,3 +1835,6 @@ const char* MR_ClusterGetMyId(){
     return clusterCtx.myId;
 }
 
+const char *MR_ClusterGetPassword(){
+    return RedisModule_GetInternalSecret ? NULL : clusterCtx.password;
+}
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -196,7 +196,7 @@ static void MR_OnStatusResponseArrived(struct redisAsyncContext* c, void* a, voi
 static void MR_OnDataResponseArrived(struct redisAsyncContext* c, void* a, void* b);  // A response to an internal-commands command
 static void MR_ConnectToShard(Node* n);
 static void MR_HelloResponseArrived(struct redisAsyncContext* c, void* a, void* b);
-static Node* MR_GetNode(const char* id);
+static Node* MR_GetNode(Cluster* cluster, const char* id);
 
 static SlotRange* NewSlotRange(uint16_t minSlot, uint16_t maxSlot) {
     SlotRange* result = MR_ALLOC(sizeof(*result));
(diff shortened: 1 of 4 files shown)


src/cluster.c
```diff
@@ -1156,34 +1131,165 @@ static int SetClusterDataShortForm(RedisModuleString** argv, int argc){
...
         }
 
         RedisModule_ClusterFreeSlotRanges(mr_staticCtx, slots);
+        if (cluster == NULL)
+            break;
     }
     RedisModule_FreeClusterNodesList(nodeList);
 
-    clusterCtx.clusterSize = mr_dictSize(clusterCtx.CurrCluster->nodes);
+    if (cluster != NULL && coveredSlots != NUMBER_OF_SLOTS) {
```
gabsow, line 1169: **Behavior narrowed for short-form CLUSTERSET.** `BuildCluster()` now returns NULL (rejecting the whole topology) unless `coveredSlots == NUMBER_OF_SLOTS` **and** no slot is claimed twice. The original `SetClusterDataShortForm()` had neither check — it accepted whatever the server reported. Because `SetClusterDataShortForm()` returns `REDISMODULE_ERR` when `BuildCluster()` is NULL, a short-form CLUSTERSET arriving during a transient window where the server reports <16384 covered slots now **fails where it used to succeed**. If this is intentional hardening, please confirm the caller (e.g. DMC) tolerates CLUSTERSET erroring during resharding / partial-coverage windows.

src/cluster.c
```diff
@@ -1156,34 +1131,165 @@ static int SetClusterDataShortForm(RedisModuleString** argv, int argc){
...
+        if (mr_listLength(n->slotRanges) > 0) {
+            SlotRange* r = mr_listNodeValue(mr_listFirst(n->slotRanges));
+            minSlot = r->minSlot;
+            maxSlot = r->maxSlot;
+        }
+        break;
+    }
+    mr_dictReleaseIterator(iter);
+    clusterCtx.minSlot = minSlot;
+    clusterCtx.maxSlot = maxSlot;
```
gabsow, line 1250: **Sign conversion makes a slotless self own every slot in the fallback path.** `minSlot`/`maxSlot` are `long long` initialised to `0`/`-1` (the documented "no hslots" sentinel). When the self node has no slot range, `maxSlot` stays `-1` and is assigned to `clusterCtx.maxSlot`, which is `size_t` → `SIZE_MAX`. `MR_ClusterIsMySlot()`'s fallback then evaluates `clusterCtx.minSlot <= slot && clusterCtx.maxSlot >= slot` → `0 <= slot && SIZE_MAX >= slot`, i.e. **true for every slot**. This fallback only runs when both `RedisModule_ClusterCanAccessKeysInSlot` and `RedisModule_ShardingGetSlotRange` are absent, so impact is limited to older builds — but on those a slotless shard wrongly claims ownership of all slots. Only set `clusterCtx.min/maxSlot` when the self node actually has a range.

src/cluster.c
```diff
@@ -1156,34 +1131,165 @@ static int SetClusterDataShortForm(RedisModuleString** argv, int argc){
...
+    }
+
+    return cluster;
+}
+
+static bool SameNode(Node* a, Node* b) {
+    return strcmp(a->id, b->id) == 0 && strcmp(a->ip, b->ip) == 0 && a->port == b->port;
+}
+
+static bool SameSlotRanges(Node* a, Node* b) {
```
gabsow, line 1184: **Order-sensitive comparison → spurious topology swap.** `SameSlotRanges()` walks both `slotRanges` lists in lockstep and returns false on the first positional mismatch, assuming identical range ordering. A node built by `SetClusterDataLongForm()`/short-form CLUSTERSET can store its ranges in a different order than the same node rebuilt by `BuildCluster()` from `RedisModule_GetClusterNodeSlotRanges()`. When the slot *set* is identical but the order differs, `SameCluster()` returns false, so `MR_UpdateClusterTopologyIfNeeded()` runs `MR_ClusterFree()` → `MR_AbortRunningExecutions()` and rebuilds — aborting in-flight executions and dropping all shard connections for a no-op change. Compare order-independently (sort ranges, or diff the `slots[]` array).

src/mr.c
```diff
@@ -1526,6 +1526,28 @@ void MR_FreeExecution(Execution* e) {
...
+    //    integrity (BuildCluster() returns non-NULL only if the topology is valid).
+    // 2. The event-loop thread (in MR_UpdateClusterTopologyIfNeeded): compare the
+    //    candidate against the installed clusterCtx.CurrCluster and swap it in
+    //    (alongside other fields, e.g., .clusterSize) only if it changed.
+    //    Since this step reads and writes the cluster context it must run on the event-loop thread.
+    //
+    // Because the comparison runs later on the event-loop thread, this function
+    // cannot return a boolean indicating whether the topology actually changed or not
+    // so it returns nothing.
+    Cluster *cluster = BuildCluster(argv, argc, password);
```
gabsow, line 1545: **Missing NULL-pointer guard on the notification path.** This `BuildCluster()` call (main-thread topology-notification path) is not protected by the `RedisModule_GetClusterNodeSlotRanges == NULL` check that `SetClusterDataShortForm()` performs before its own `BuildCluster()` call. Inside `BuildCluster()`, `RedisModule_GetClusterNodeSlotRanges(...)` is invoked directly and only its *return value* is asserted non-NULL — the function pointer itself is never checked. `SetClusterDataShortForm()` guards this precisely because some Redis builds don't export the symbol ("OSS Redis without the backport"). If a build exposes `RedisModuleEvent_ClusterTopologyChange` (so the callback fires) but not `RedisModule_GetClusterNodeSlotRanges`, this path calls through a NULL function pointer → crash. Add the same `RedisModule_GetClusterNodeSlotRanges == NULL` guard here (or at the top of `BuildCluster()`).

gabsow, line 1545: **M1 (must-fix): heap OOB write + leak on the OSS auto-refresh path (`argc==0`).**

This calls `BuildCluster(NULL, 0, …)` → `InitClusterData(cluster, NULL, 0)`, where (`cluster.c:996-998`):
```c
clusterSetCommand = MR_ALLOC(sizeof(char*) * argc);  // argc==0 -> MR_ALLOC(0)
clusterSetCommandSize = argc;                        // 0
clusterSetCommand[0] = MR_STRDUP(CLUSTER_SET_FROM_SHARD_COMMAND); // write [0] of a 0-len alloc
```
Two problems, both on the feature's primary path (fires on every topology event):
1. **OOB write** to `clusterSetCommand[0]`. Survives shipped builds (zmalloc floors to 8B) but is a deterministic Invalid-write under the `-DVALGRIND` build (there `MR_ALLOC=malloc`) → Valgrind/ASan CI failure.
2. **Leak**: `FreeCluster`'s free loop is bounded by `clusterSetCommandSize==0`, so the strdup at `[0]` is never freed — ~32 bytes/event, worst on the common no-op-skip branch.

Fix: special-case `argc==0` in `InitClusterData` → `clusterSetCommand=NULL; clusterSetCommandSize=0;` and skip the `[0]` write (mirrors `MR_RefreshClusterData` at `cluster.c:892-893`). `CopyClusterSetArgs`/`SetMyId` already tolerate `argc==0`.

src/cluster.c
```diff
@@ -1156,34 +1131,165 @@ static int SetClusterDataShortForm(RedisModuleString** argv, int argc){
...
+    }
+    mr_dictReleaseIterator(iter);
+    return entry == NULL;
+}
+
+void MR_UpdateClusterTopologyIfNeeded(void* ctx){
+    Cluster* cluster = ctx;
+    RedisModule_Assert(cluster != NULL);
+
+    if (SameCluster(cluster, clusterCtx.CurrCluster)) {
```
gabsow, line 1223: **m1 (minor): this dedup is dead on the CLUSTERSET path.** `MR_SetClusterData` (`cluster.c:1362-1363`) calls `MR_ClusterFree()` unconditionally *before* the candidate is built/compared, so on the CLUSTERSET path `clusterCtx.CurrCluster` is already `NULL` here → `SameCluster(x, NULL)` is always false → always installs (and the abort already happened). So the no-op skip only protects the OSS event path; an identical CLUSTERSET re-send (the MOD-16399 DMC reconnect case) still tears down + aborts in-flight. Fine if intended — MOD-16399 has its own fix — just worth stating so nobody assumes #105 subsumes it.

src/cluster.c
```diff
@@ -1090,49 +1096,18 @@ static int ParseShardEntry(RedisModuleString** argv, int argc, int index,
...
-        password = RedisModule_StringPtrLen(argv[2], NULL);
-        break;
-    }
-    default:
-        RedisModule_Assert(0);
-    }
-
+Cluster* BuildCluster(RedisModuleString** argv, int argc, const char* password) {
     size_t numNodes;
     char **nodeList = RedisModule_GetClusterNodesList(mr_staticCtx, &numNodes);
```
gabsow, line 1101: **m2 (minor, pre-existing): unlocked cluster-introspection off the main thread.** `BuildCluster` calls `RedisModule_GetClusterNodesList` / `GetClusterNodeInfo` / `GetClusterNodeSlotRanges` with no `RedisModule_ThreadSafeContextLock`. On the new OSS path that's fine (main thread, GIL held), but the CLUSTERSET path reaches `BuildCluster` on the **event-loop** thread (`MR_EventLoopAddTask` → `MR_ClusterSetFromCommand`). `MR_RefreshClusterData` locks the identical call (`cluster.c:939-941`). Lifted verbatim from the old `SetClusterDataShortForm` so not a regression — worth a follow-up to wrap these (guard against double-lock on the main-thread OSS path).

src/cluster.c
```diff
@@ -1156,34 +1131,165 @@ static int SetClusterDataShortForm(RedisModuleString** argv, int argc){
...
+    Cluster* cluster = ctx;
+    RedisModule_Assert(cluster != NULL);
+
+    if (SameCluster(cluster, clusterCtx.CurrCluster)) {
+        FreeCluster(cluster);
+        return;
+    }
+
+    if (clusterCtx.CurrCluster)
+        MR_ClusterFree();
```
gabsow, line 1229: **m3 (design trade-off): every real change does a full teardown.** `SameNode` compares id/ip/port only, so a pure slot-move reshard (same master set) makes `SameSlotRanges` differ → `SameCluster` false → this `MR_ClusterFree()` aborts all in-flight multi-shard executions, drops every inter-shard connection, and regenerates the run id — on every reshard. Correctness holds (clients retry) and it isn't new behavior (manual REFRESHCLUSTER did the same), just newly automatic. The in-place slot-repoint alternative preserves connections on slot-only moves. Recommend documenting this as accepted per-reshard churn, with a follow-up to evaluate in-place if it hurts at scale.

src/cluster.h
```diff
@@ -27,10 +27,16 @@ void MR_ClusterCopyAndSendMsgBySlot(size_t slot, functionId function, char* msg,
 
 functionId MR_ClusterRegisterMsgReceiver(MR_ClusterMessageReceiver receiver);
 
+typedef struct Cluster Cluster;
+Cluster* BuildCluster(RedisModuleString** argv, int argc, const char* password);
```
gabsow, line 31: **n3 (nit): naming.** `BuildCluster` is the only non-`MR_`-prefixed export in this header, and LibMR is statically linked into the module. Suggest `MR_BuildCluster` for namespace hygiene (here + `cluster.c:1099` + call sites `cluster.c:1288`, `mr.c:1545`).

src/mr.c
```diff
@@ -1526,6 +1526,28 @@ void MR_FreeExecution(Execution* e) {
     MR_FREE(e);
 }
 
+void MR_UpdateClusterTopology() {
+    RedisModuleString** argv = NULL;
```
gabsow, line 1530: **n4 (nit):** `argv`/`argc` are always `NULL`/`0` here — could inline `BuildCluster(NULL, 0, password)` with a short comment, or keep for symmetry with the CLUSTERSET call. Taste.

gabsow (commented):
**Automated deep review** (adversarial multi-agent pass + manual cross-check), read one commit at a time as you suggested — 9 topology commits `83eb39b..6f9ed9a`.

**Verdict: mergeable with one must-fix (M1 below).** The split is clean and correct: I traced every writer of `clusterCtx.CurrCluster` and they all serialize on the event-loop thread, so no lock is needed and the design comment in `mr.c` is accurate. Integrity validation (overlap + coverage) is a nice touch.

- **M1 (must-fix)** — inline on `mr.c`: `InitClusterData` OOB write + per-event leak on the `argc==0` OSS path.
- **m1/m2/m3 (minor)** — inline: dedup dead on the CLUSTERSET path (MOD-16399 not covered here); unlocked cluster-introspection on the event-loop CLUSTERSET path (pre-existing); swap-on-every-change aborts in-flight + drops connections per reshard (the design trade-off).
- **n3/n4 (nits)** — inline: naming + dead locals.

**Test gap (not inline):** the 9 topology commits add **no LibMR unit tests** — `BuildCluster`, `SameCluster`, `MR_UpdateClusterTopology`, and both integrity-rejection branches (overlap `cluster.c:1150-1157`, gap `cluster.c:1169-1174`) are all unexercised. A small `MRTESTS.UPDATETOPOLOGY`-style command driven under ASan would have caught M1 immediately. Strongly recommend before merge.

gabsow (approved):

src/cluster.c
galcohen-redislabs: Indeed, the vast majority of the module apis are not thread-safe (GetClusterNodesList, GetClusterNodeInfo and GetClusterNodeSlotRanges included).
This is a real risk that is much wider than the scope of this ticket.
I opened a new ticket for that: https://redislabs.atlassian.net/browse/MOD-16950

src/cluster.c
galcohen-redislabs: This is intentional - the original validations were too permissive.

src/cluster.c
galcohen-redislabs: The order is coming from `RM_GetClusterNodeSlotRanges()` where the order is guaranteed. Then the order is preserved since we we call `mr_listAddNodeTail(slotRanges)`.
The long-form clusterset might indeed pose a problem because it might not preserve the order (we need to verify with the dmc team), but there's another ticket for that: https://redislabs.atlassian.net/browse/MOD-16399

src/cluster.c
galcohen-redislabs: Intended. We will handle the long-form case in https://redislabs.atlassian.net/browse/MOD-16399

src/cluster.c
galcohen-redislabs: Yes, intended. Slightly safer this way.
If we see any performance degradation we should reconsider this tradeoff.

src/cluster.c
galcohen-redislabs: The clusterCtx.min/maxSlot are legacy (they are a fallback for the case of some missing APIs) so will never be used. We will remove them at some point.

src/cluster.h
galcohen-redislabs: Makes sense. Will push it in some future commit.

src/mr.c
galcohen-redislabs: argc changed to 1 to align with the fact that the command-name is always expected.
inlining will make it less readable because `(argv, argc, ...)` is more readable than `(NULL, 1, ...)`

src/mr.c
galcohen-redislabs: We get to that path only through the new topology-change notifications which are newer than said API, so therefore we can safely assume it is not null in this case.

galcohen-redislabs: Problem gone when argc was changed to 1.

PR #108, opened 2026-07-16
MOD-16956 Fix execution leak when an initiator execution is aborted after completing

gabsow:
## What

Fixes a memory leak of a whole multi-shard execution (the `Execution` struct plus all its accumulated result records) that LeakSanitizer catches in the `oss_cluster` sanitizer CI leg once topology auto-refresh (MOD-16382) starts aborting in-flight executions on every topology change.

Caught by the `linux-sanitizer / x64-jammy-oss_cluster` leg on RedisTimeSeries#2116: [failing job](https://github.com/RedisTimeSeries/RedisTimeSeries/actions/runs/29515712405/job/87680155101). All tests pass there — the failure is purely the LeakSanitizer report at process exit (`Direct leak … MR_ExecutionAlloc ← MR_CreateExecution ← TSDB_mrange_MR`, plus ~440 KB of parsed series/chunks/labels hanging off it).

## Root cause — a race between completion and abort

1. `MR_RunExecution` completes an internal-protocol **initiator** execution: it fires the done callback, sets `e->callbacks.done.callback = NULL`, and defers `MR_DeleteExecution` to the event loop.
2. Before that runs, a topology change triggers `MR_AbortRunningExecutions`, which removes the (already-completed) execution from `executionsDict` and schedules `MR_ExecutionAbortedOnClusterChange`.
3. That task saw `done.callback == NULL` and `return`ed **before** `MR_FreeExecution()`, so the run-owned reference was never released. Its "terminal task" treatment in the dispatcher then stranded the still-queued `MR_DisposeExecution`. Net: `refCount` stuck at 1 → the execution and its results leak.

## Fix (abort path only — 2 functions)

- `MR_ExecutionAbortedOnClusterChange` now **always** calls `MR_FreeExecution()`, invoking the done callback only if it has not already fired.
- `MR_DeleteExecution` only disposes when it actually removed the execution from the dictionary (`mr_dictDelete(...) == DICT_OK`); if the abort already claimed it, it does nothing.

All dictionary operations run on the event-loop thread, so exactly one path claims each execution → exactly one `MR_FreeExecution()`: no leak, no double-free, no use-after-free. The two changes are coupled — always-free is only safe once the competing delete is gated. (The idle-timeout path has the same latent shape but is not the observed leak; left untouched.)

## Verification

- **Functional:** RTS pinned to this branch, `tests/flow/test_asm.py` 7/7 on `oss-cluster` (incl. the auto-refresh tests that churn topology during multi-shard queries) — no crash, double-free, or hang.
- **ASAN:** re-running the exact CI sanitizer leg (`flow-linux.yml`, x64/jammy, `run_sanitizer=true`) against a RTS branch pinned to this fix to confirm the `LeakSanitizer: detected` report is gone (LeakSanitizer is Linux-only, hence CI rather than local). The abort-path change is what fixes the leak.

---

> [!NOTE]
> **Medium Risk**
> Changes multi-shard execution lifecycle and refcount teardown on cluster abort and idle timeout; incorrect handling could leak or double-free `Execution` objects.
> 
> **Overview**
> Fixes a **LeakSanitizer** leak when a multi-shard initiator execution finishes (done callback already ran, `MR_DisposeExecution` queued) and **cluster topology abort** schedules `MR_ExecutionAbortedOnClusterChange` on the same execution.
> 
> **Execution task callbacks** now return `bool`: `true` means the callback freed the execution and `MR_ExecutionMain` must stop touching it; `false` means keep draining the task queue.
> 
> On **idle timeout** and **cluster abort**, if the done callback was already cleared by a racing normal completion, the handler **does not** call `MR_FreeExecution` (avoids double-free) and returns **false** so a queued `MR_DisposeExecution` still runs. When the handler owns teardown, it frees and returns **true**.
> 
> `MR_ExecutionMain` uses that return value instead of comparing callback pointers, and **does not re-arm the idle timer** after a terminal timeout/abort task that returned false (prevents repeated timeout errors); it still processes any remaining tasks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2612c871f49fff128c296d6d29d61116517611e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/mr.c
+++ b/src/mr.c
@@ -1365,6 +1365,15 @@ static void MR_ExecutionMain(void* pd) {
         return;
     }
 
+    // A terminal handler (idle-timeout / cluster-abort) that returned false did not
+    // dispose e, but it has already removed e from executionsDict. We still drain any
+    // task queued behind it (e.g. a MR_DisposeExecution from a racing normal completion),
+    // but we must NOT re-arm the idle timer for such an execution: nothing can cancel it
+    // anymore, so it would re-fire forever (appending to e->errors every tick). A pending
+    // MR_DisposeExecution, if one was queued, still runs and frees e.
+    bool terminal = (callback == MR_ExecutionTimedOutInternal ||
+                     callback == MR_ExecutionAbortedOnClusterChange);
+
     pthread_mutex_lock(&e->eLock);
     /* pop current task out */
     mr_listDelNode(e->tasks, head);
@@ -1374,7 +1383,7 @@ static void MR_ExecutionMain(void* pd) {
         /* more work to do, for fairness we will not run now.
          * We will add ourselves to the thread pool */
         mr_thpool_add_work(mrCtx.executionsThreadPool, MR_ExecutionMain, e);
-    } else {
+    } else if (!terminal) {
         e->timeoutTask = MR_EventLoopAddTaskWithDelay(MR_ExecutionTimedOut, e, e->timeoutMS);
     }
 
--- a/src/mr.c
+++ b/src/mr.c
@@ -984,7 +984,11 @@ static void MR_DisposeExecution(Execution* e, void* pd) {
  * executions dictionary and send a dispose task */
 static void MR_DeleteExecution(void* ctx) {
     Execution* e = ctx;
-    mr_dictDelete(mrCtx.executionsDict, e->id);
+    /* If it's already gone, an abort-on-cluster-change claimed it and will dispose it; don't
+     * schedule a second terminal task (that would double-free). */
+    if (mr_dictDelete(mrCtx.executionsDict, e->id) != DICT_OK) {
+        return;
+    }
     /* Send dispose execution task, this will be last task this execution will ever receive. */
     MR_ExecutionAddTask(e, MR_DisposeExecution, NULL);
 }
@@ -1318,10 +1322,14 @@ static void MR_ExecutionTimedOut(void* ctx) {
 
 // Thread pool task: fires the done callback with a cluster topology change error
 static void MR_ExecutionAbortedOnClusterChange(Execution* e, void* pd) {
-    e->errors = array_append(e->errors, MR_ErrorRecordCreate("cluster topology changed"));
-    if (!MR_ExecutionInvokeCallback(e, &e->callbacks.done))
-        return;
-    e->callbacks.done.callback = NULL; /* make sure the done callback will not be called again */
+    /* The execution may have already completed (done already fired) and be awaiting its queued
+     * MR_DeleteExecution. Fire done only if still pending, but always free: we claimed it out of
+     * the dictionary so this is its sole terminal task. */
+    if (e->callbacks.done.callback) {
+        e->errors = array_append(e->errors, MR_ErrorRecordCreate("cluster topology changed"));
+        MR_ExecutionInvokeCallback(e, &e->callbacks.done);
+        e->callbacks.done.callback = NULL;
+    }
     MR_FreeExecution(e);
 }
 
--- a/src/mr.c
+++ b/src/mr.c
@@ -57,11 +57,14 @@ typedef struct RemoteFunctionDef {
 
 typedef struct Step Step;
 
-typedef void (*ExecutionTaskCallback)(Execution* e, void* pd);
+// Returns true if the callback disposed (freed) the execution, in which case the
+// caller (MR_ExecutionMain) must not touch it afterwards. Returns false otherwise,
+// so the caller keeps draining the execution's task queue.
+typedef bool (*ExecutionTaskCallback)(Execution* e, void* pd);
 
 /* functions declarations */
 static void MR_ExecutionAddTask(Execution* e, ExecutionTaskCallback callback, void* pd);
-static void MR_RunExecution(Execution* e, void* pd);
+static bool MR_RunExecution(Execution* e, void* pd);
 static Record* MR_RunStep(Execution* e, Step* s);
 void MR_FreeExecution(Execution* e);
 
@@ -609,7 +612,7 @@ static size_t MR_PerformStepDoneOp(Execution* e, size_t stepIndex) {
 }
 
 /* Execution task */
-static void MR_StepDone(Execution* e, void* pd) {
+static bool MR_StepDone(Execution* e, void* pd) {
     RedisModuleString* payload = pd;
 
     /* deserialize record, set it on the right step. */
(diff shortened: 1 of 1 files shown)


src/mr.c
```diff
@@ -1318,10 +1322,14 @@ static void MR_ExecutionTimedOut(void* ctx) {
...
 // Thread pool task: fires the done callback with a cluster topology change error
 static void MR_ExecutionAbortedOnClusterChange(Execution* e, void* pd) {
-    e->errors = array_append(e->errors, MR_ErrorRecordCreate("cluster topology changed"));
-    if (!MR_ExecutionInvokeCallback(e, &e->callbacks.done))
-        return;
-    e->callbacks.done.callback = NULL; /* make sure the done callback will not be called again */
+    /* The execution may have already completed (done already fired) and be awaiting its queued
+     * MR_DeleteExecution. Fire done only if still pending, but always free: we claimed it out of
+     * the dictionary so this is its sole terminal task. */
+    if (e->callbacks.done.callback) {
```
galcohen-redislabs: How is this check more bulletproof than the `if (!MR_ExecutionInvokeCallback(...))` that is used now?

src/mr.c
```diff
@@ -984,7 +984,11 @@ static void MR_DisposeExecution(Execution* e, void* pd) {
  * executions dictionary and send a dispose task */
 static void MR_DeleteExecution(void* ctx) {
     Execution* e = ctx;
-    mr_dictDelete(mrCtx.executionsDict, e->id);
+    /* If it's already gone, an abort-on-cluster-change claimed it and will dispose it; don't
+     * schedule a second terminal task (that would double-free). */
```
galcohen-redislabs: Did we actually hit a double free? If so - how? which threads raced exactly?

src/mr.c
gabsow: It isn't more bulletproof at detecting the null — `MR_ExecutionInvokeCallback` returns false in exactly the same case (`!callback->callback`). The bug was never the check, it's the early `return`: the old

```c
if (!MR_ExecutionInvokeCallback(e, &e->callbacks.done)) return;
...
MR_FreeExecution(e);
```

returns *before* `MR_FreeExecution()` when the execution already fired its done callback (completed, then aborted) → the run-owned reference is never released → leak. My version just guarantees the free always runs; the `if (e->callbacks.done.callback)` only avoids re-invoking / appending a pointless error on an already-completed execution.

If you'd rather keep the original idiom, the minimal equivalent is to just drop the `return` (the callback no-ops on NULL):

```c
MR_ExecutionInvokeCallback(e, &e->callbacks.done);
e->callbacks.done.callback = NULL;
MR_FreeExecution(e);
```

Happy to switch to whichever reads better.

src/mr.c
gabsow: No — what we hit is a **leak**, not a double free. This guard is what *prevents* a double free that the abort-side change would otherwise introduce.

refCount is 1 after the initiator's `MR_Run` + `MR_FreeExecution`; exactly one terminal task must drop it to 0. On the race two terminal tasks get scheduled for the same execution:

1. `MR_RunExecution` completes an internal-protocol initiator exec (thread pool) → fires done, NULLs `done.callback`, and defers `MR_DeleteExecution` (event loop).
2. A topology change → `MR_AbortRunningExecutions` (event loop) removes it from the dict + schedules `MR_ExecutionAbortedOnClusterChange` (thread pool).

`MR_DeleteExecution` and `MR_AbortRunningExecutions` both run on the event-loop thread (serialized), so there's no race over *who* removes it from the dict. But once the abort task always calls `MR_FreeExecution()`, both terminal tasks would free → refCount 1→0→-1 → double free / UAF. This guard makes `MR_DeleteExecution` a no-op when it didn't actually remove the exec (the abort already claimed it), so exactly one terminal task frees. Dict removal is the single-owner token.

(In the *current* code there's no double free precisely because the abort early-returns without freeing — which is exactly why it leaks.)

gabsow:
Pushed a hardening commit (6952e8d) that reworks the fix. The previous approach (abort **always** frees, `MR_DeleteExecution` gated on `mr_dictDelete == DICT_OK`) has two problems:

1. **Use-after-free.** When the abort claims the execution out of the dictionary, the `MR_DeleteExecution` that a racing normal completion already queued still runs and dereferences `e` (`mr_dictDelete(e->id)`) on the **event-loop thread**, while `MR_ExecutionAbortedOnClusterChange` frees `e` on a **thread-pool thread** — an unsynchronized heap-use-after-free that would surface in the same ASAN leg.
2. It left the idle-timeout twin (`MR_ExecutionTimedOutInternal`) with the same latent leak.

New model — single, race-free disposal:
- `ExecutionTaskCallback` returns whether it disposed the execution; `MR_ExecutionMain` skips post-processing only when it actually freed, otherwise it **drains the task queue** so a pending `MR_DisposeExecution` still runs (stranding that dispose was the original leak).
- The abort/idle-timeout handlers free **only when they own termination** (done callback still pending). When the done callback already fired, they return `false` and **defer to the queued `MR_DeleteExecution → MR_DisposeExecution`** — which keeps `e` alive for that event-loop dereference.
- `MR_DeleteExecution` goes back to always scheduling the dispose (the dict-ownership gate is no longer needed and would re-introduce the leak on the defer path).

Result: exactly one `MR_FreeExecution` per execution, no thread frees `e` while another still dereferences it — no leak, no double-free, no UAF — and it now covers both the topology-change abort and the idle-timeout paths.

(Supersedes #112, which took the same defer+drain approach off `master`; closing that in favor of this PR.)

gabsow:
Follow-up review — two issues on the general **non-internal / non-Local** distributed-initiator path. The RedisTimeSeries (internal-command) path is unaffected and passes the oss_cluster sanitizer.

1. **Regression — fixed in 2612c87.** After a terminal handler (`MR_ExecutionAbortedOnClusterChange` / `MR_ExecutionTimedOutInternal`) returned `false`, `MR_ExecutionMain` re-armed the idle timer on an execution already removed from `executionsDict` → nothing could cancel it, so it re-fired every `timeoutMS` forever, appending a fresh `"execution max idle reached"` to `e->errors` each tick. Now it re-arms only for non-terminal callbacks.

2. **Pre-existing leak — not fixed here.** A non-`Local`/non-internal **initiator** completes (`MR_RunExecution` nulls `done.callback`) without queuing `MR_DeleteExecution`, and waits for peers' `NOTIFY_DONE`. If an abort/idle-timeout removes it from the dict first, the terminal handler returns `false` without freeing and nothing else disposes it → the `Execution` leaks once. This predates the branch (base `a6d14e8` leaked here too); left for the `MR_UpdateClusterTopology` refactor (MOD-16956). With (1) fixed it no longer spins — just the one-time leak.

galcohen-redislabs (approved):

PR #109, opened 2026-07-17
MOD-16382 Some logs and cleanups

galcohen-redislabs:
> [!NOTE]
> **Medium Risk**
> Changes cluster configuration paths and threading around topology swaps; mis-handling the topology-events vs legacy-command interaction could leave cluster state stale or ignore needed fallbacks.
> 
> **Overview**
> Adds a **`topologyEvents`** flag through **`MR_Init`** / **`mr_init`** so modules can prefer Redis topology notifications over manual cluster commands.
> 
> When topology events are enabled and at least one notification was handled (`MR_TopologyEventSeen`), **REFRESHCLUSTER** and **CLUSTERSET** (short/long) are **skipped** with warnings; if events are enabled but none arrived yet, those commands still run as a fallback. **`MR_UpdateClusterTopology`** sets the seen flag and schedules updates via **`MR_UpdateClusterTopologyIfNeededUnderLock`** (thread-safe context lock on the event loop).
> 
> **`BuildCluster`** is renamed **`MR_BuildCluster`** with clearer NULL handling on incomplete slot coverage. **`INFOCLUSTER`** node replies use a **variable-length** array and emit **all** slot ranges per node (not a single range). Tests read node **status** from **`n[-1]`** instead of a fixed index; the test module initializes with **`topology_events: true`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57b1c2a002fb9a6727bd83e0a2702d4398001eec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -165,7 +165,10 @@ struct ClusterCtx {
     int isOss;
     functionId networkTestMsgReceiver;
     char *password;
+    bool topologyEvents;
 }clusterCtx;
+
+bool MR_TopologyEventSeen = false;
 
 typedef struct ClusterSetCtx {
     RedisModuleBlockedClient* bc;
@@ -871,11 +874,22 @@ static Node* MR_CreateNode(Cluster* cluster, const char* id, const char* ip, uns
 }
 
 static void MR_RefreshClusterData(){
+    if (clusterCtx.topologyEvents) {
+        if (MR_TopologyEventSeen) {
+            RedisModule_Log(mr_staticCtx, "warning", "Got unexpected cluster refresh command when topology events are enabled - skipping");
+            return;
+        }
+        RedisModule_Log(mr_staticCtx, "warning", "Got unexpected cluster refresh command when topology events are enabled");
+        // Topology events are enabled, but we haven't seen one yet. maybe the redis-server doesn't support them.
+        // So to be on the safe side, we'll process the command .
+    } else {
+        RedisModule_Assert(!MR_TopologyEventSeen);  // Since no handler was registered for topology events
+        RedisModule_Log(mr_staticCtx, "notice", "Got cluster refresh command");
+    }
+
     if(clusterCtx.CurrCluster){
         MR_ClusterFree();
     }
-
-    RedisModule_Log(mr_staticCtx, "notice", "Got cluster refresh command");
 
     if(!(RedisModule_GetContextFlags(mr_staticCtx) & REDISMODULE_CTX_FLAGS_CLUSTER)){
         return;
@@ -1259,7 +1273,18 @@ void MR_UpdateClusterTopologyIfNeeded(void* ctx){
 }
 
 static int SetClusterDataShortForm(RedisModuleString** argv, int argc){
-    RedisModule_Log(mr_staticCtx, "notice", "Got cluster set command (short form)");
+    if (clusterCtx.topologyEvents) {
+        if (MR_TopologyEventSeen) {
+            RedisModule_Log(mr_staticCtx, "warning", "Got unexpected cluster set command (short form) when topology events are enabled - skipping");
+            return REDISMODULE_OK;
+        }
+        RedisModule_Log(mr_staticCtx, "warning", "Got unexpected cluster set command (short form) when topology events are enabled");
+        // Topology events are enabled, but we haven't seen one yet. maybe the redis-server doesn't support them.
+        // So to be on the safe side, we'll process the command .
+    } else {
+        RedisModule_Assert(!MR_TopologyEventSeen);  // Since no handler was registered for topology events
+        RedisModule_Log(mr_staticCtx, "notice", "Got cluster set command (short form)");
+    }
 
     // RedisModule_GetClusterNodeSlotRanges may be NULL when the host Redis
     // build does not export it (e.g. OSS Redis without the backport). Reject
@@ -1299,7 +1324,18 @@ static int SetClusterDataShortForm(RedisModuleString** argv, int argc){
 }
 
 static void SetClusterDataLongForm(RedisModuleString** argv, int argc){
-    RedisModule_Log(mr_staticCtx, "notice", "Got cluster set command (long form)");
+    if (clusterCtx.topologyEvents) {
+        if (MR_TopologyEventSeen) {
+            RedisModule_Log(mr_staticCtx, "warning", "Got unexpected cluster set command (long form) when topology events are enabled - skipping");
+            return;
+        }
+        RedisModule_Log(mr_staticCtx, "warning", "Got unexpected cluster set command (long form) when topology events are enabled");
+        // Topology events are enabled, but we haven't seen one yet. maybe the redis-server doesn't support them.
+        // So to be on the safe side, we'll process the command .
+    } else {
+        RedisModule_Assert(!MR_TopologyEventSeen);  // Since no handler was registered for topology events
+        RedisModule_Log(mr_staticCtx, "notice", "Got cluster set command (long form)");
+    }
 
     clusterCtx.CurrCluster = MR_CALLOC(1, sizeof(*clusterCtx.CurrCluster));
     InitClusterData(clusterCtx.CurrCluster, argv, argc);
@@ -1775,7 +1811,7 @@ static void MR_NetworkTest(RedisModuleCtx *ctx, const char *sender_id, uint8_t t
     RedisModule_Log(ctx, "notice", "got a nextwork test msg");
 }
 
-int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
+int MR_ClusterInit(RedisModuleCtx* rctx, char *password, bool topologyEvents) {
     clusterCtx.CurrCluster = NULL;
     clusterCtx.callbacks = array_new(MR_ClusterMessageReceiver, 10);
     clusterCtx.nodesMsgIds = mr_dictCreate(&mr_dictTypeHeapStrings, NULL);
@@ -1784,6 +1820,7 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
     clusterCtx.clusterSize = 1;
     clusterCtx.isOss = true;
     clusterCtx.password = password ? MR_STRDUP(password) : NULL;
+    clusterCtx.topologyEvents = topologyEvents;
     memset(clusterCtx.myId, '0', REDISMODULE_NODE_ID_LEN);
 
     /* Note: RedisModule_GetContextFlags() does NOT report
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1613,19 +1613,24 @@ static void MR_ClusterInfo(void* pd) {
     mr_dictEntry *entry = NULL;
     while((entry = mr_dictNext(iter))){
         Node* n = mr_dictGetVal(entry);
-        RedisModule_ReplyWithArray(ctx, 18);
+        RedisModule_ReplyWithArray(ctx, REDISMODULE_POSTPONED_LEN);
+        long nodeLen = 0;
         RedisModule_ReplyWithStringBuffer(ctx, "id", strlen("id"));
         RedisModule_ReplyWithStringBuffer(ctx, n->id, strlen(n->id));
+        nodeLen += 2;
         RedisModule_ReplyWithStringBuffer(ctx, "ip", strlen("ip"));
         RedisModule_ReplyWithStringBuffer(ctx, n->ip, strlen(n->ip));
+        nodeLen += 2;
         RedisModule_ReplyWithStringBuffer(ctx, "port", strlen("port"));
         RedisModule_ReplyWithLongLong(ctx, n->port);
+        nodeLen += 2;
         RedisModule_ReplyWithStringBuffer(ctx, "unixSocket", strlen("unixSocket"));
         if(n->unixSocket){
             RedisModule_ReplyWithStringBuffer(ctx, n->unixSocket, strlen(n->unixSocket));
         }else{
             RedisModule_ReplyWithStringBuffer(ctx, "None", strlen("None"));
         }
+        nodeLen += 2;
         RedisModule_ReplyWithStringBuffer(ctx, "runid", strlen("runid"));
         if(n->runId){
             RedisModule_ReplyWithStringBuffer(ctx, n->runId, strlen(n->runId));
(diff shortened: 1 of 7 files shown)


gabsow (commented):
Reviewed — clean cleanup, LGTM. Checked the three things that could bite:

- **INFOCLUSTER `REDISMODULE_POSTPONED_LEN` counter**: `nodeLen` matches the emitted elements for every node shape (18 for a single range, +4 per extra range; correct with `runId`/`unixSocket` NULL and with 0 ranges). The `common.py` `n[17]`→`n[-1]` change correctly follows the now-variable length.
- **`BuildCluster`→`MR_BuildCluster` NULL refactor**: behavior-identical in all paths (no double-free, no leak, same return/log); all call sites + the header updated.
- **Dropping `RedisModule_Assert(n->slotRanges->len == 1)`**: 👍 that assert was a crash risk for multi-range nodes.

Nit: stale doc comment at `src/cluster.c:555` (`MR_RefreshClusterData / BuildCluster`) still uses the old name — outside this diff, so noting here.

(Context, not a request: this is cleanups only — it doesn't address the #108 Execution-abort leak or the `mr.c:939` InternalCommand assert; those are separate.)

gabsow (dismissed):

gabsow (commented):
Ci of libMR is red

gabsow (commented):
Re-reviewing the new commit `e61f88ca` (*"gate topology-related commands on topology-events config"*) — it wasn't in the head I looked at before (`6bdc724f`), so flagging that my earlier approval/LGTM was on the old head and doesn't cover this. The gating logic itself reads fine; three things:

**1. `MR_Init` is a public API break — this is what's reding RTS #2121.** Adding `bool topologyEvents` to `MR_Init` / `MR_ClusterInit` breaks every consumer not updated in lockstep: RTS's `register_mr` (`libmr_integration.c:1372`) still passes 3 args, so #2121 fails to compile on **every** leg. Worth landing this together with the consumer call-site bump (or at least calling it out) so a pin bump doesn't red the whole matrix.

**2. The gating is frozen at load, but `ts-topology-events` is a mutable config.** `clusterCtx.topologyEvents` is captured once in `MR_ClusterInit`, and the consumer subscribes the topology-change handler only at load (RTS `module.c`, based on the load-time config). But the config is registered *without* `REDISMODULE_CONFIG_IMMUTABLE` and has a live setter — so a runtime `CONFIG SET ts-topology-events ...` updates the config yet silently changes **nothing** in the gating or the subscription. Either mark it boottime/immutable, or read the live value / (un)subscribe on change.

**3. `MR_TopologyEventSeen` is an unsynchronized cross-thread `bool`.** Written on the main thread (`MR_UpdateClusterTopology`, from the server-event callback) and read on the event-loop thread (`MR_RefreshClusterData`, `SetClusterDataShortForm/LongForm`, which are deferred via `MR_EventLoopAddTask`). Benign on common arches, but this exact main-vs-event-loop sharing is where the HELLO UAF lived — consider an atomic, or a comment that it's an intentional single-writer/benign race.

Minor: the `RedisModule_Assert(!MR_TopologyEventSeen)` in the events-disabled branch is safe *today* (only reachable when the load-time flag was false → the handler was never subscribed → the flag stays false), but it's coupled to that load-time-only assumption; if you address #2 it'll need revisiting. And please confirm the enterprise/DMC `CLUSTERSET` path still works with events defaulting on — it currently relies on the "no event seen yet → process" fallback.

gabsow:
### The `gate topology-related commands…` commit probably belongs in its own PR

Following up on the re-review: the `gate topology-related commands on topology-events config` commit (the `bool topologyEvents` `MR_Init` param + `MR_TopologyEventSeen` gating) is the one reding the `build` legs here — the `MR_Init` signature change breaks consumers not updated in lockstep: RTS `register_mr` (fixed in RedisTimeSeries#2121) **and LibMR's own Rust test binding** (`tests/mr_test_module`), which is the current clippy failure:
```
error[E0061]: this function takes 4 arguments but 3 arguments were supplied
  MR_Init(...)  →  could not compile `lib_mr`
```

I also checked whether the RTS #2121 tests exercise this gating — they don't (no test fires a manual `REFRESHCLUSTER`/`CLUSTERSET` while events are on; setup takes the "none seen yet → process" fallback). So it's a standalone production-behavior change riding along in a "logs and cleanups" PR.

**Suggestion:** split it into its own PR with the Rust binding + RTS call-site updated together (and the frozen-config / cross-thread-`bool` points addressed). That gets #109 back to the clean "logs + INFOCLUSTER multi-range" scope I originally LGTM'd (`6bdc724f`), unbreaks its build, and lets the gating behavior get reviewed on its own merits. If the gating isn't a hard requirement, dropping it is even simpler.

galcohen-redislabs:
> Reviewed — clean cleanup, LGTM. Checked the three things that could bite:
> 
> * **INFOCLUSTER `REDISMODULE_POSTPONED_LEN` counter**: `nodeLen` matches the emitted elements for every node shape (18 for a single range, +4 per extra range; correct with `runId`/`unixSocket` NULL and with 0 ranges). The `common.py` `n[17]`→`n[-1]` change correctly follows the now-variable length.
> * **`BuildCluster`→`MR_BuildCluster` NULL refactor**: behavior-identical in all paths (no double-free, no leak, same return/log); all call sites + the header updated.
> * **Dropping `RedisModule_Assert(n->slotRanges->len == 1)`**: 👍 that assert was a crash risk for multi-range nodes.
> 
> Nit: stale doc comment at `src/cluster.c:555` (`MR_RefreshClusterData / BuildCluster`) still uses the old name — outside this diff, so noting here.
> 
> (Context, not a request: this is cleanups only — it doesn't address the #108 Execution-abort leak or the `mr.c:939` InternalCommand assert; those are separate.)

Don't see any stale "BuildBCluster" in comments.

> Re-reviewing the new commit `e61f88ca` (_"gate topology-related commands on topology-events config"_) — it wasn't in the head I looked at before (`6bdc724f`), so flagging that my earlier approval/LGTM was on the old head and doesn't cover this. The gating logic itself reads fine; three things:
> 
> **1. `MR_Init` is a public API break — this is what's reding RTS #2121.** Adding `bool topologyEvents` to `MR_Init` / `MR_ClusterInit` breaks every consumer not updated in lockstep: RTS's `register_mr` (`libmr_integration.c:1372`) still passes 3 args, so #2121 fails to compile on **every** leg. Worth landing this together with the consumer call-site bump (or at least calling it out) so a pin bump doesn't red the whole matrix.
> 
> **2. The gating is frozen at load, but `ts-topology-events` is a mutable config.** `clusterCtx.topologyEvents` is captured once in `MR_ClusterInit`, and the consumer subscribes the topology-change handler only at load (RTS `module.c`, based on the load-time config). But the config is registered _without_ `REDISMODULE_CONFIG_IMMUTABLE` and has a live setter — so a runtime `CONFIG SET ts-topology-events ...` updates the config yet silently changes **nothing** in the gating or the subscription. Either mark it boottime/immutable, or read the live value / (un)subscribe on change.
> 
> **3. `MR_TopologyEventSeen` is an unsynchronized cross-thread `bool`.** Written on the main thread (`MR_UpdateClusterTopology`, from the server-event callback) and read on the event-loop thread (`MR_RefreshClusterData`, `SetClusterDataShortForm/LongForm`, which are deferred via `MR_EventLoopAddTask`). Benign on common arches, but this exact main-vs-event-loop sharing is where the HELLO UAF lived — consider an atomic, or a comment that it's an intentional single-writer/benign race.
> 
> Minor: the `RedisModule_Assert(!MR_TopologyEventSeen)` in the events-disabled branch is safe _today_ (only reachable when the load-time flag was false → the handler was never subscribed → the flag stays false), but it's coupled to that load-time-only assumption; if you address #2 it'll need revisiting. And please confirm the enterprise/DMC `CLUSTERSET` path still works with events defaulting on — it currently relies on the "no event seen yet → process" fallback.

The config is immutable, hence the assert.

galcohen-redislabs:
> ### The `gate topology-related commands…` commit probably belongs in its own PR
> Following up on the re-review: the `gate topology-related commands on topology-events config` commit (the `bool topologyEvents` `MR_Init` param + `MR_TopologyEventSeen` gating) is the one reding the `build` legs here — the `MR_Init` signature change breaks consumers not updated in lockstep: RTS `register_mr` (fixed in RedisTimeSeries#2121) **and LibMR's own Rust test binding** (`tests/mr_test_module`), which is the current clippy failure:
> 
> ```
> error[E0061]: this function takes 4 arguments but 3 arguments were supplied
>   MR_Init(...)  →  could not compile `lib_mr`
> ```
> 
> I also checked whether the RTS #2121 tests exercise this gating — they don't (no test fires a manual `REFRESHCLUSTER`/`CLUSTERSET` while events are on; setup takes the "none seen yet → process" fallback). So it's a standalone production-behavior change riding along in a "logs and cleanups" PR.
> 
> **Suggestion:** split it into its own PR with the Rust binding + RTS call-site updated together (and the frozen-config / cross-thread-`bool` points addressed). That gets #109 back to the clean "logs + INFOCLUSTER multi-range" scope I originally LGTM'd (`6bdc724f`), unbreaks its build, and lets the gating behavior get reviewed on its own merits. If the gating isn't a hard requirement, dropping it is even simpler.

The Rust bindings are part of the repo (even though I don't think anyone uses them), so when breaking internal APIs we need to break the rust ones as well. Therefore I don't see why this deserves a seperate PR.

src/cluster.c
```diff
@@ -1251,8 +1272,25 @@ void MR_UpdateClusterTopologyIfNeeded(void* ctx){
     mr_dictEmpty(clusterCtx.nodesMsgIds, NULL);
 }
 
+void MR_UpdateClusterTopologyIfNeededUnderLock(void* ctx){
```
gabsow, line 1275: This is a **partial** fix for the same HELLO use-after-free that #113 addresses, and it overlaps #113.

Taking the GIL here makes the **event-driven** swap mutually exclusive with main-thread `MR_ClusterHello` (which holds the GIL) → the UAF in the repro (freed via `MR_UpdateClusterTopologyIfNeeded`) is closed. But only the event path is scheduled under this wrapper (`mr.c:1549`). `MR_RefreshClusterData` (REFRESHCLUSTER) and `SetClusterDataShortForm/LongForm` (CLUSTERSET) still `MR_ClusterFree()` / reassign `clusterCtx.CurrCluster` on the EL thread **without** the GIL, and `MR_ClusterHello` still reads `CurrCluster->runId` synchronously — so a HELLO can still UAF via those paths. #113 (defer the read to the EL) covers all free paths.

⇒ keep #113 as the complete fix; this lock is either redundant (with #113) or insufficient (without it). Also please double-check this can't deadlock: nothing in the wrapped path (`MR_ClusterFree` → reconnect → `SendAuthCommandIfNeeded`, which itself calls `RedisModule_ThreadSafeContextLock`) may re-enter the non-recursive lock on this thread.

src/cluster.c
```diff
@@ -1766,6 +1826,7 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
     clusterCtx.clusterSize = 1;
     clusterCtx.isOss = true;
     clusterCtx.password = password ? MR_STRDUP(password) : NULL;
+    clusterCtx.topologyEvents = topologyEvents;
```
gabsow, line 1829: `clusterCtx.topologyEvents` is frozen here at init. Combined with the load-time-only subscription on the consumer side, and given `ts-topology-events` is registered *mutable* (no `REDISMODULE_CONFIG_IMMUTABLE`), a runtime `CONFIG SET ts-topology-events ...` updates the config but changes nothing here. Mark the config boottime/immutable, or read the live value / (un)subscribe on change.

gabsow (commented):
Re-review at `f6937515` (+96/-29). The mechanical work is clean — `BuildCluster→MR_BuildCluster` + the NULL refactor are behavior-preserving, the INFOCLUSTER postponed-length/multi-range counter is correct for every node shape (and dropping the `slotRanges->len==1` assert removes a real crash risk), and the `MR_Init` consumers (incl. the Rust binding) are updated so builds are green again.

The one that matters is the new **`f6937515` "atomic swap under lock"** commit — see the inline note. Short version: it's a *second, partial* fix for the same HELLO use-after-free that #113 fixes, and it only covers the topology-**event** path. So it overlaps #113 and shouldn't replace it.

Other points:
- **Frozen config:** `clusterCtx.topologyEvents` is captured once in `MR_ClusterInit` and the topology-change subscription is load-time only, but `ts-topology-events` is registered *mutable* (no `IMMUTABLE`) → a runtime `CONFIG SET` silently no-ops. Mark it boottime/immutable, or wire runtime changes through.
- **`MR_TopologyEventSeen` is an unsynchronized cross-thread `bool`** — written on the main thread (`MR_UpdateClusterTopology`), read on the EL thread (`MR_RefreshClusterData`/`SetClusterData*`); the new lock doesn't cover this flag.
- **Scope:** the gating, the `MR_Init` signature change, and the atomic-swap are three functional changes (one overlapping #113) inside a "logs and cleanups" PR — worth splitting the topology-behavior commits out and keeping this PR to the rename / INFOCLUSTER / logs.

gabsow:
Consolidating my earlier comments to cut the noise:

- **`BuildCluster` nit** — you're right, it's gone at the current head (all `MR_BuildCluster` now). Resolved, ignore it.
- **Scope split** — fair, your call; dropping that.
- **Config (minor / FYI)** — one factual note: `ts-topology-events` isn't actually marked immutable — it's registered `REDISMODULE_CONFIG_UNPREFIXED` (no `REDISMODULE_CONFIG_IMMUTABLE`) with a live setter (RTS `config.c:356`), so `CONFIG SET` is accepted at runtime. The assert is still safe — but only because LibMR's `clusterCtx.topologyEvents` and the event subscription are both frozen at load, so a runtime `CONFIG SET` silently no-ops. If the intent is "immutable", worth actually adding `REDISMODULE_CONFIG_IMMUTABLE` so it doesn't advertise a setting that does nothing. Low priority.

The one thing still worth a decision (from my latest review): the new `f6937515` "atomic swap under lock" only guards the topology-**event** path. `MR_RefreshClusterData` (REFRESHCLUSTER) and `SetClusterData*` (CLUSTERSET) still free/reassign `CurrCluster` on the EL thread without the GIL, and `MR_ClusterHello` still reads it synchronously — so a HELLO can still UAF via those paths. It overlaps #113, which (deferring the read to the EL) covers all of them. Worth picking one rather than shipping both half-overlapping.

galcohen-redislabs:
> Re-review at `f6937515` (+96/-29). The mechanical work is clean — `BuildCluster→MR_BuildCluster` + the NULL refactor are behavior-preserving, the INFOCLUSTER postponed-length/multi-range counter is correct for every node shape (and dropping the `slotRanges->len==1` assert removes a real crash risk), and the `MR_Init` consumers (incl. the Rust binding) are updated so builds are green again.
> 
> The one that matters is the new **`f6937515` "atomic swap under lock"** commit — see the inline note. Short version: it's a _second, partial_ fix for the same HELLO use-after-free that #113 fixes, and it only covers the topology-**event** path. So it overlaps #113 and shouldn't replace it.

I'm not convinced this is the case. Need to see the test that has UAF with the latest version (which will include PR108 on the LibMR part, latest code of this branch on the ts part, and then check with and without 113).

> 
> Other points:
> 
> * **Frozen config:** `clusterCtx.topologyEvents` is captured once in `MR_ClusterInit` and the topology-change subscription is load-time only, but `ts-topology-events` is registered _mutable_ (no `IMMUTABLE`) → a runtime `CONFIG SET` silently no-ops. Mark it boottime/immutable, or wire runtime changes through.

It is registered with `REDISMODULE_CONFIG_IMMUTABLE` in config.c:468 

> * **`MR_TopologyEventSeen` is an unsynchronized cross-thread `bool`** — written on the main thread (`MR_UpdateClusterTopology`), read on the EL thread (`MR_RefreshClusterData`/`SetClusterData*`); the new lock doesn't cover this flag.

Theoretically this is true, but the behavior is fine with the slight chance of race on this bool: what might happen is that w CLUSTERSET will get served even though it should not be if setting to true during the event handler was missed. We can live with that.

> * **Scope:** the gating, the `MR_Init` signature change, and the atomic-swap are three functional changes (one overlapping #113) inside a "logs and cleanups" PR — worth splitting the topology-behavior commits out and keeping this PR to the rename / INFOCLUSTER / logs.

gabsow:
Now that #108 is merged to master and #113 (the HELLO use-after-free + ClusterSetCtx leak fix) is approved, the `f6937515` "atomic swap under lock" here is **superseded — you can drop it**:

- #113 fixes the HELLO UAF completely by deferring `MR_ClusterHello`'s `CurrCluster->runId` read to the event loop, so it's serialized against **all** free paths. The lock here only guards the topology-**event** free path — `MR_RefreshClusterData` (REFRESHCLUSTER) and `SetClusterData*` (CLUSTERSET) still race — so it's a partial mitigation of the same bug.

For the record, #109 does **not** cover #113 (so #113 still has to land regardless): here `MR_ClusterHello` (`cluster.c:1507`) still reads synchronously, and `MR_ClusterSetInternal` (`cluster.c:1464`) still passes `NULL` for `free_privdata` — so the ClusterSetCtx leak is untouched.

Dropping the lock gets #109 back to the clean logs / INFOCLUSTER-multi-range / gating scope, and #113 owns the UAF + ClusterSetCtx fixes.

src/cluster.c
gabsow: thas becouse of the conflict between old and new way of config set - we need the old why for runtime chenging config

galcohen-redislabs:
Done cleaner in PR #117

PR #112, opened 2026-07-17
MOD-16382 Fix Execution leak on topology-change abort / idle-timeout race

gabsow:
## Summary

Initiator `TS.MRANGE`/`TS.QUERYINDEX` **Executions** (and all their parsed result Records) leak when a cluster **topology change** aborts in-flight executions. Surfaced by the topology-change flow tests in [RedisTimeSeries#2116](https://github.com/RedisTimeSeries/RedisTimeSeries/pull/2116) — the `linux-sanitizer / x64-jammy-oss_cluster` job reports:

```
Direct leak of 1632 byte(s) in 6 object(s):
    MR_ExecutionAlloc  src/mr.c:516
    MR_CreateExecution src/mr.c:542
    TSDB_mrange_MR     ... (initiator, internal-command fan-out)
Indirect leak of 2.45 MB in 599 object(s): parsed Series/chunks/labels off e->results
SUMMARY: AddressSanitizer: 2762418 byte(s) leaked in 7973 allocation(s).
```

The 6 `Execution` objects are the **only** direct-leak root; everything else is indirect, hanging off their `results` arrays.

## Root cause

`MR_ExecutionAbortedOnClusterChange` and `MR_ExecutionTimedOutInternal` **early-return without `MR_FreeExecution`** when the done callback was already fired by a racing normal completion (`done.callback == NULL`). But `MR_ExecutionMain` unconditionally treated those two callbacks (alongside `MR_DisposeExecution`) as "the execution was disposed" and returned **without popping the task or rescheduling**.

`MR_RunExecution` always nulls `done.callback` *before* it queues `MR_DeleteExecution` → `MR_DisposeExecution`. So whenever a `MR_DisposeExecution` is pending, `done.callback` is already NULL — exactly the condition that makes the abort/timeout handler take its no-free branch.

Interleaving (EL = event loop, TP = thread pool), all shards have replied:

1. **[EL]** last shard reply → results parsed, completion gate holds → schedule `MR_RunExecution`. `e->tasks=[RUN]`.
2. **[EL]** topology change → `MR_ClusterFree` → `MR_AbortRunningExecutions`: cancels idle timer, `dictDelete`s `e`, appends abort. `e->tasks=[RUN, ABORT]`.
3. **[TP]** `MR_RunExecution` fires done, **nulls it**, schedules `MR_DeleteExecution`; RUN popped, runner reschedules. `e->tasks=[ABORT]`.
4. **[EL]** `MR_DeleteExecution`: dict delete is a no-op, appends `MR_DisposeExecution`. `e->tasks=[ABORT, DISPOSE]`.
5. **[TP]** `MR_ExecutionAbortedOnClusterChange` → `done==NULL` → early-return (no free) → `MR_ExecutionMain` returns **without popping/rescheduling**.

Terminal state: `e->tasks=[ABORT, DISPOSE]` frozen, `e` out of the dict, idle timer gone, refCount stuck at 1. `MR_DisposeExecution` never runs → the Execution and its results leak permanently (survives to shutdown LeakSanitizer).

This has been latent since **MOD-14439** (abort in-flight executions on topology change) but only became reachable once **MOD-16382** began installing topology from `ClusterTopologyChange` notifications, which runs `MR_ClusterFree → MR_AbortRunningExecutions` on **every** real topology change.

## Fix

Make `ExecutionTaskCallback` return whether it disposed (freed) the execution. `MR_ExecutionMain` skips post-processing only when the callback actually disposed it; otherwise it drains the task queue as before, so a pending `MR_DisposeExecution` still runs. The abort/idle-timeout handlers return `false` on the early-return path (they did not free) — fixing the leak while preserving a single free.

**Why not just always-free in the handlers** (an earlier approach): on the current code that is a **use-after-free**. When the abort runs *before* the already-scheduled `MR_DeleteExecution`, freeing in the handler leaves `MR_DeleteExecution` operating on freed memory. Deferring to the queued `MR_DisposeExecution` keeps exactly one free and no double-free.

## Testing

- `make -C src/` builds cleanly (no new warnings).
- Verified against the ASAN signature above; the fix targets the exact stranded-`MR_DisposeExecution` path. The RTS#2116 `linux-sanitizer / x64-jammy-oss_cluster` job is the end-to-end reproducer — after bumping the RTS LibMR pin to this commit it should go green.

---

> [!NOTE]
> **Medium Risk**
> Touches execution lifecycle and freeing on cluster abort/timeout race paths in core map-reduce threading; logic is narrow but memory-safety critical.
> 
> **Overview**
> Fixes **Execution** and result **Record** leaks when cluster topology change or idle timeout races normal completion on initiator map-reduce runs.
> 
> **`ExecutionTaskCallback`** now returns whether the task **actually freed** the execution. **`MR_ExecutionMain`** only stops without popping/rescheduling when that return is **true** (e.g. **`MR_DisposeExecution`**), instead of treating abort/timeout handlers the same even when they did not free.
> 
> **`MR_ExecutionAbortedOnClusterChange`** and **`MR_ExecutionTimedOutInternal`** return **`false`** when the done callback was already cleared by a racing **`MR_RunExecution`**, so the runner keeps draining the queue and a queued **`MR_DisposeExecution`** still runs—avoiding stranded tasks and leaks without double-freeing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc32e5488ef55b89a511ac4ddd8a1b9b7cf6b86c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/mr.c
+++ b/src/mr.c
@@ -57,11 +57,14 @@ typedef struct RemoteFunctionDef {
 
 typedef struct Step Step;
 
-typedef void (*ExecutionTaskCallback)(Execution* e, void* pd);
+// Returns true if the callback disposed (freed) the execution, in which case the
+// caller (MR_ExecutionMain) must not touch it afterwards. Returns false otherwise,
+// so the caller keeps draining the execution's task queue.
+typedef bool (*ExecutionTaskCallback)(Execution* e, void* pd);
 
 /* functions declarations */
 static void MR_ExecutionAddTask(Execution* e, ExecutionTaskCallback callback, void* pd);
-static void MR_RunExecution(Execution* e, void* pd);
+static bool MR_RunExecution(Execution* e, void* pd);
 static Record* MR_RunStep(Execution* e, Step* s);
 void MR_FreeExecution(Execution* e);
 
@@ -609,7 +612,7 @@ static size_t MR_PerformStepDoneOp(Execution* e, size_t stepIndex) {
 }
 
 /* Execution task */
-static void MR_StepDone(Execution* e, void* pd) {
+static bool MR_StepDone(Execution* e, void* pd) {
     RedisModuleString* payload = pd;
 
     /* deserialize record, set it on the right step. */
@@ -636,10 +639,11 @@ static void MR_StepDone(Execution* e, void* pd) {
         /* All shards are done running the step, we can continue the execution. */
         MR_RunExecution(e, NULL);
     }
+    return false;  // never disposes the execution here
 }
 
 /* Execution task */
-static void MR_SetRecord(Execution* e, void* pd) {
+static bool MR_SetRecord(Execution* e, void* pd) {
     RedisModuleString* payload = pd;
 
     /* deserialize record, set it on the right step. */
@@ -668,6 +672,7 @@ static void MR_SetRecord(Execution* e, void* pd) {
         /* There is enough records to process, lets continue running. */
         MR_RunExecution(e, NULL);
     }
+    return false;  // never disposes the execution here
 }
 
 /* Remote function call, runs on the event loop */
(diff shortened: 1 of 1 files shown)


gabsow:
Superseded by #108, which now carries the hardened version of this exact fix (defer-to-dispose + task-queue drain, covering both the topology-change abort and the idle-timeout twin, with no use-after-free). Consolidating on #108.

PR #116, opened 2026-07-26
MOD-16600 Serialize legacy topology replacement

gabsow:
## Summary

Serialize legacy `CLUSTERSET` and `REFRESHCLUSTER` topology replacement with Redis' main thread so command execution cannot observe partially rebuilt `clusterCtx`.

Jira: [MOD-16600](https://redislabs.atlassian.net/browse/MOD-16600)

Related topology-event work: [LibMR #109](https://github.com/RedisGears/LibMR/pull/109)

## Support-package evidence

The MOD-16600 support package contains four Redis crashes at `LibMR/src/mr.c`, where `MR_RunStep()` asserts because a worker received `StepType_InternalCommand`.

The failing Enterprise ASM workload:

- scales a multi-shard database while coordinated TimeSeries queries are running;
- uses `ts-libmr-protocol INTERNAL`, not the legacy GEARS protocol;
- repeatedly installs topology through long-form `CLUSTERSET`;
- crashes with `MR_RunStep -> MR_RunExecution -> MR_ExecutionMain`.

## Root cause

Legacy topology replacement runs on the LibMR event-loop thread:

1. `MR_ClusterFree()` clears `CurrCluster` and temporarily sets `clusterCtx.clusterSize = 1`.
2. The replacement path rebuilds the global `clusterCtx` in place.
3. Only after construction finishes is the real cluster size restored.

Redis' main thread concurrently reads `clusterCtx.clusterSize` from `MR_CreateExecution()`. During that free/rebuild window, a TimeSeries internal-command execution can observe a false standalone state and be marked `ExecutionFlag_Local`. Its worker later reaches the intentional `StepType_InternalCommand` assertion.

`REFRESHCLUSTER` performs the same global free/rebuild sequence through `MR_RefreshClusterData()`, so it requires the same synchronization boundary.

## Fix

- `MR_ClusterSetFromCommand()` holds the Redis thread-safe-context lock across the complete `MR_SetClusterData()` operation.
- `MR_ClusterRefreshFromCommand()` holds the lock across the complete `MR_RefreshClusterData()` operation.
- The two old inner locks around `RedisModule_Call()` and `RedisModule_GetClusterNodeInfo()` are removed because the entire refresh now runs under the outer lock; keeping them would recursively acquire a non-recursive lock.

The protection around those Redis APIs is therefore widened, not removed.

The fix does not change `MR_CreateExecution()`, add an internal-command routing exception, or add steady-state execution locking. The native topology-event path is separate and is covered by LibMR #109.

## Deterministic regressions

The test module registers a real supported internal command using `MR_ExecutionBuilderInternalCommand()`. Both regressions run on real two- and three-shard OSS clusters and contain no test-only production hook:

- `CLUSTERSET`: one connection installs a valid long-form topology while another creates the internal-command execution. Repeated valid ranges widen the real free/rebuild window.
- `REFRESHCLUSTER`: a barrier starts 50 real refreshes concurrently with 500 real internal commands, then verifies that the node remains responsive and distributed execution still reaches every shard.

### A/B proof

On the baseline code, the long-form `CLUSTERSET` regression deterministically crashes Redis:

```text
mr.c:944 false is not true
MR_RunStep
MR_RunExecution
MR_ExecutionMain
```

The new `REFRESHCLUSTER` regression was also transplanted onto and rebuilt from the pre-fix commit (`e0ba9d4`). It failed on both two and three shards with:

```text
ResponseError('internal command execution failed')
```

Redis command statistics record the concurrent internal command as failed while all 50 topology refresh commands succeed. With the outer refresh lock, the identical test passes on both cluster widths.

## Validation

- Focused two-shard concurrent `CLUSTERSET` regression: pass.
- Focused three-shard concurrent `CLUSTERSET` regression: pass.
- Focused pre-fix two-shard `REFRESHCLUSTER` A/B: expected failure.
- Focused pre-fix three-shard `REFRESHCLUSTER` A/B: expected failure.
- Focused fixed two-shard `REFRESHCLUSTER` A/B: pass.
- Focused fixed three-shard `REFRESHCLUSTER` A/B: pass.
- Standalone matrix: 36 tests, 0 failures.
- One-shard OSS cluster matrix: 36 tests, 0 failures.
- Two-shard OSS cluster matrix: 36 tests, 0 failures.
- Three-shard OSS cluster matrix: 36 tests, 0 failures.
- Release build, Python compilation, and `git diff --check`: pass.

The lock is used only during topology replacement, so there is no steady-state throughput cost. Local control-path measurements are recorded in [this PR comment](https://github.com/RedisGears/LibMR/pull/116#issuecomment-5083127195).

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 42d59e08bd6eac9f00b3cc45fadf81ddc1ef95db. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1382,12 +1382,16 @@ static void MR_ClusterRefreshFromCommand(void* ctx){
     RedisModule_UnblockClient(bc, NULL);
 }
 
-/* runs in the event loop so its safe to update cluster
- * topology here */
+/* Runs on the LibMR event loop. Redis' main thread also reads clusterCtx while
+ * creating executions, so hold the Redis lock across the legacy free/rebuild
+ * sequence and do not expose its temporary one-shard state. */
 static void MR_ClusterSetFromCommand(void* ctx){
     ClusterSetCtx* csCtx = ctx;
     if (!clusterCtx.CurrCluster || csCtx->force) {
-        if (MR_SetClusterData(csCtx->argv, csCtx->argc) != REDISMODULE_OK) {
+        RedisModule_ThreadSafeContextLock(mr_staticCtx);
+        int status = MR_SetClusterData(csCtx->argv, csCtx->argc);
+        RedisModule_ThreadSafeContextUnlock(mr_staticCtx);
+        if (status != REDISMODULE_OK) {
             csCtx->errReply = CLUSTER_ERROR" Failed to set cluster topology";
         }
     }
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -896,9 +896,7 @@ static void MR_RefreshClusterData(){
     memcpy(clusterCtx.myId, clusterCtx.CurrCluster->myId, REDISMODULE_NODE_ID_LEN + 1);
     clusterCtx.CurrCluster->nodes = mr_dictCreate(&mr_dictTypeHeapStrings, NULL);
 
-    RedisModule_ThreadSafeContextLock(mr_staticCtx);
     RedisModuleCallReply *allSlotsReply = RedisModule_Call(mr_staticCtx, "cluster", "c", "slots");
-    RedisModule_ThreadSafeContextUnlock(mr_staticCtx);
 
     RedisModule_Assert(RedisModule_CallReplyType(allSlotsReply) == REDISMODULE_REPLY_ARRAY);
     for(size_t i = 0 ; i < RedisModule_CallReplyLength(allSlotsReply) ; ++i){
@@ -934,10 +932,7 @@ static void MR_RefreshClusterData(){
         // invoking `cluster slot` from RM_Call will always return the none tls port.
         // For for information refer to: https://github.com/redis/redis/pull/12233
         int port = 0;
-        RedisModule_ThreadSafeContextLock(mr_staticCtx);
         RedisModule_GetClusterNodeInfo(mr_staticCtx, nodeId, NULL, NULL, &port, NULL);
-        RedisModule_ThreadSafeContextUnlock(mr_staticCtx);
-
 
         Node* n = MR_GetNode(clusterCtx.CurrCluster, nodeId);
         if(!n){
@@ -1371,11 +1366,13 @@ static int MR_SetClusterData(RedisModuleString** argv, int argc){
     }
 }
 
-/* runs in the event loop so its safe to update cluster
- * topology here */
+/* Runs on the LibMR event loop. Hold the Redis lock across the complete
+ * free/rebuild sequence so the main thread cannot observe partial topology. */
 static void MR_ClusterRefreshFromCommand(void* ctx){
     RedisModuleBlockedClient* bc = ctx;
+    RedisModule_ThreadSafeContextLock(mr_staticCtx);
     MR_RefreshClusterData();
+    RedisModule_ThreadSafeContextUnlock(mr_staticCtx);
     RedisModuleCtx* rCtx = RedisModule_GetThreadSafeContext(bc);
     RedisModule_ReplyWithSimpleString(rCtx, "OK");
     RedisModule_FreeThreadSafeContext(rCtx);
--- a/src/mr.c
+++ b/src/mr.c
@@ -562,9 +562,7 @@ Execution* MR_CreateExecution(ExecutionBuilder* builder, MRError** err) {
     }
 
     e->flags |= ExecutionFlag_Initiator;
-    /* Internal commands must always be distributed through the event loop.
-     * During a topology replacement MR_ClusterFree temporarily reports one shard. */
-    if (!MR_ClusterIsClusterMode() && !MR_IsInternalCommandsExecution(e)) {
+    if (!MR_ClusterIsClusterMode()) {
         e->flags |= ExecutionFlag_Local;
     }
 
--- a/src/mr.c
+++ b/src/mr.c
@@ -562,7 +562,9 @@ Execution* MR_CreateExecution(ExecutionBuilder* builder, MRError** err) {
     }
 
     e->flags |= ExecutionFlag_Initiator;
-    if (!MR_ClusterIsClusterMode()) {
+    /* Internal commands must always be distributed through the event loop.
+     * During a topology replacement MR_ClusterFree temporarily reports one shard. */
+    if (!MR_ClusterIsClusterMode() && !MR_IsInternalCommandsExecution(e)) {
         e->flags |= ExecutionFlag_Local;
     }
 
--- a/tests/mr_test_module/pytests/test_basic.py
+++ b/tests/mr_test_module/pytests/test_basic.py
@@ -1,6 +1,7 @@
 import pytest
 from common import MRTestDecorator, ShardsConnectionTimeoutException, initialiseCluster, TimeLimit
 import time
+import unittest
 
 @MRTestDecorator()
 def testBasicMR(env, conn):
@@ -91,6 +92,14 @@ def testRemoteTaskOnAllShards(env, conn):
     for i in range(100):
         conn.execute_command('del', 'doc%d' % i)
     env.expect('lmrtest.dbsize').equal(0)
+
+@MRTestDecorator()
+def testInternalCommandExecutionUsesEventLoop(env, conn):
+    if not env.isCluster():
+        raise unittest.SkipTest()
+    # Internal-command executions use the event-loop distribution path even
+    # when the current topology contains only one shard.
+    env.expect('lmrtest.internalcommand').equal(env.shardsCount)
 
 @MRTestDecorator(skipOnVersionLowerThan='8.0.0', skipOnCluster=False)
 def testInternalCommandsAreNotAllowed(env, conn):
--- a/tests/mr_test_module/pytests/test_basic.py
+++ b/tests/mr_test_module/pytests/test_basic.py
@@ -94,11 +94,12 @@ def testRemoteTaskOnAllShards(env, conn):
     env.expect('lmrtest.dbsize').equal(0)
 
 @MRTestDecorator()
-def testInternalCommandExecutionUsesEventLoop(env, conn):
+def testInternalCommandIsNeverRoutedLocally(env, conn):
     if not env.isCluster():
         raise unittest.SkipTest()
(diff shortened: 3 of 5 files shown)


src/mr.c
```diff
@@ -562,7 +562,9 @@ Execution* MR_CreateExecution(ExecutionBuilder* builder, MRError** err) {
     }
 
     e->flags |= ExecutionFlag_Initiator;
-    if (!MR_ClusterIsClusterMode()) {
+    /* Internal commands must always be distributed through the event loop.
+     * During a topology replacement MR_ClusterFree temporarily reports one shard. */
+    if (!MR_ClusterIsClusterMode() && !MR_IsInternalCommandsExecution(e)) {
```
gabsow: MR_IsInternalCommandsExecution is a config, and it will be allways 1

gabsow: @galcohen-redislabs have seen the same issue somewere else

gabsow: Correction after checking the exact Enterprise test and support package: `MR_IsInternalCommandsExecution(e)` is a per-execution step-type check, not the `ts-libmr-protocol` config. The failing ASM test and shard logs both show `INTERNAL`, not GEARS. Commit a7c7aa9 adds a real `MR_ExecutionBuilderInternalCommand` regression: without the guard a one-shard cluster crashes at the same `MR_RunStep` assertion; with the guard it passes on 1/2/3 shards and TLS. So this guard protects the supported internal-command path.

src/cluster.c
```diff
@@ -1382,12 +1382,16 @@ static void MR_ClusterRefreshFromCommand(void* ctx){
...
-/* runs in the event loop so its safe to update cluster
- * topology here */
+/* Runs on the LibMR event loop. Redis' main thread also reads clusterCtx while
+ * creating executions, so hold the Redis lock across the legacy free/rebuild
+ * sequence and do not expose its temporary one-shard state. */
 static void MR_ClusterSetFromCommand(void* ctx){
     ClusterSetCtx* csCtx = ctx;
     if (!clusterCtx.CurrCluster || csCtx->force) {
-        if (MR_SetClusterData(csCtx->argv, csCtx->argc) != REDISMODULE_OK) {
+        RedisModule_ThreadSafeContextLock(mr_staticCtx);
```
TalBarYakar, line 1388: maybe we should have like a define that let you execute safe context execution between 
RedisModule_ThreadSafeContextLock(mr_staticCtx);
CODE ...
RedisModule_ThreadSafeContextUnlock(mr_staticCtx);

so unlock will never be forgatten 

```suggestion
        EXECUTE_SAFE_CONTEXT
        ({
            int status = MR_SetClusterData(csCtx->argv, csCtx->argc);
        })
```

src/cluster.c
```diff
@@ -1382,12 +1382,16 @@ static void MR_ClusterRefreshFromCommand(void* ctx){
     RedisModule_UnblockClient(bc, NULL);
```
TalBarYakar, line 1379: Should we lock this path too? REFRESHCLUSTER still does the same free/rebuild flow, so Redis can still see clusterSize = 1 while creating an internal-command execution.

inside the MR_RefreshClusterData we have lock unlock but it only for the api and not for all 
can you check it ?

src/cluster.c
gabsow: I agree that a reusable lock helper could prevent mismatched pairs, but I do not think it belongs in this bug-fix PR. The proposed statement-expression macro is also compiler-specific, and `status` as written would be scoped inside the macro block and unavailable afterward. I kept the explicit lock/unlock pattern used elsewhere in LibMR; a general helper can be considered as separate cleanup.

src/cluster.c
```diff
@@ -1382,12 +1382,16 @@ static void MR_ClusterRefreshFromCommand(void* ctx){
...
+/* Runs on the LibMR event loop. Redis' main thread also reads clusterCtx while
+ * creating executions, so hold the Redis lock across the legacy free/rebuild
+ * sequence and do not expose its temporary one-shard state. */
 static void MR_ClusterSetFromCommand(void* ctx){
     ClusterSetCtx* csCtx = ctx;
     if (!clusterCtx.CurrCluster || csCtx->force) {
-        if (MR_SetClusterData(csCtx->argv, csCtx->argc) != REDISMODULE_OK) {
+        RedisModule_ThreadSafeContextLock(mr_staticCtx);
+        int status = MR_SetClusterData(csCtx->argv, csCtx->argc);
+        RedisModule_ThreadSafeContextUnlock(mr_staticCtx);
```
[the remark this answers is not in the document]
gabsow: Thanks. I intentionally did not restore that guard. `MR_IsInternalCommandsExecution()` is topology-independent, so the guard also changes true standalone behavior: with `CurrCluster == NULL`, the distribution path drops the message and the execution times out. MOD-16600 is the multi-shard topology-replacement race shown by the support package. The PR now prevents the Redis main thread from observing the transient one-shard state in both `CLUSTERSET` and `REFRESHCLUSTER`, while leaving `MR_CreateExecution()` unchanged. Supporting a manually-created internal execution on a static one-shard topology is a separate pre-existing behavior and should be handled explicitly rather than through this race fix.

src/cluster.c
gabsow: Good catch — yes, this path had the same observable free/rebuild window. I moved the lock outward to cover the complete `MR_RefreshClusterData()` operation and removed its two nested per-API lock pairs, so there is no recursive lock. `MR_RefreshClusterData()` has only this caller. I verified standalone initialization, one-shard cluster initialization, explicit two-shard `REFRESHCLUSTER`, and the concurrent two-shard `CLUSTERSET` regression.

gabsow:
Performance check (local microbenchmark; benchmark code was not committed): 100 realistic long-form `CLUSTERSET` replacements on an eight-shard OSS cluster while a second client continuously issued `PING`.

| metric | unlocked | locked | delta |
|---|---:|---:|---:|
| replacement median | 0.189 ms | 0.199 ms | +0.010 ms |
| replacement p95 | 0.248 ms | 0.264 ms | +0.016 ms |
| concurrent `PING` median | 0.067 ms | 0.100 ms | +0.033 ms |
| concurrent `PING` p95 | 0.112 ms | 0.180 ms | +0.068 ms |
| concurrent `PING` max | 0.858 ms | 1.855 ms | +0.997 ms |

This lock is used only during topology replacement, so there is no steady-state throughput cost. The observed control-path pause is small on an idle local cluster. Caveat: `MR_AbortRunningExecutions()` is inside the protected replacement, so a heavily loaded Enterprise shard with many active executions may have a longer tail pause. The real ASM scale workload is the appropriate follow-up measurement for a production tail-latency bound; if that is unacceptable, the larger optimization would be to build a complete candidate topology off-lock and lock only the abort/swap phase.

src/cluster.c
gabsow: @TalBarYakar One scope question before we keep this change: the MOD-16600 support package exercises Enterprise long-form `CLUSTERSET`; `REFRESHCLUSTER` is an OSS/manual path and is not required to fix the reported incident. It does have the identical `MR_ClusterFree()` free/rebuild race, and the current follow-up protects it with the same outer lock and passes the targeted tests. Would you prefer keeping this same-class hardening in PR #116, or reverting the `REFRESHCLUSTER` part and tracking it separately to keep this PR strictly scoped to the observed path?

gabsow:
Addressed the lock-scope concern in `d1b9a61`.

Long-form `CLUSTERSET` now constructs a complete candidate topology without touching `clusterCtx`; the Redis lock covers only the final `MR_ClusterFree()` plus publication of all replacement fields. This removes the range-parsing/build cost from the main-thread pause while preserving the MOD-16600 synchronization boundary.

I also removed the `REFRESHCLUSTER` hardening from this PR. That command was not present in the support-package failure path, and narrowing its lock correctly would require a separate Redis-API snapshot/build refactor.

Validation after the refactor:

- deterministic race passes twice on both 2 and 3 shards;
- standalone and 1/2/3-shard matrices each pass 35/35;
- release build, formatting, Python compilation, and diff checks pass.

gabsow:
Scope update: I reverted both the candidate-topology refactor and `2d994b4` (`REFRESHCLUSTER` serialization).

The final production diff is intentionally limited to `MR_ClusterSetFromCommand()`: lock around the existing `MR_SetClusterData()` call. This fixes the exact long-form `CLUSTERSET` free/rebuild race found in the MOD-16600 support package without changing topology construction, `REFRESHCLUSTER`, or native topology events.

The deterministic internal-command reproduction passes on both 2 and 3 shards on this final tree. Tal's build-off-to-the-side suggestion remains a possible control-path latency optimization, but it is not required for this bug fix.

gabsow:
Decision update after the `clusterCtx` ownership discussion: the branch was reset to `2d994b4`.

The three later commits (`d1b9a61`, `aec48ee`, and `6559195`) were removed from branch history. The PR again protects both complete legacy mutation paths:

- `MR_SetClusterData()` for `CLUSTERSET`;
- `MR_RefreshClusterData()` for `REFRESHCLUSTER`.

The refresh API calls remain protected by the outer lock; their old inner lock pairs are removed only to avoid recursive locking.

gabsow:
Added direct `REFRESHCLUSTER` race coverage in [`42d59e0`](https://github.com/RedisGears/LibMR/commit/42d59e08bd6eac9f00b3cc45fadf81ddc1ef95db).

The test starts 50 real `MRTESTS.REFRESHCLUSTER` calls and 500 real registered internal commands from separate connections behind the same barrier. It uses no delay or test hook in production code. After the race it also verifies that Redis is responsive and that distributed work still reaches every shard.

Clean A/B result with the identical test:

| build | 2 shards | 3 shards |
|---|---|---|
| pre-fix `e0ba9d4` | fails: `internal command execution failed` | fails: `internal command execution failed` |
| fixed `42d59e0` | passes | passes |

The pre-fix Redis stats show all 50 refreshes succeeding while the concurrent internal command is recorded as failed. The complete fixed 2- and 3-shard matrices also pass 36/36 each.

TalBarYakar (approved):

PR #118, opened 2026-07-26
MOD-17016 Detect Enterprise-managed native clusters

gabsow:
## Summary

- Detect Enterprise-managed native clusters from the configured `clusterplugin.cluster-topology-uri` module config.
- Preserve the behavior from #98 for a private Redis binary running an unmanaged OSS cluster.
- Select the Enterprise `_proxy-filtered` transport in ASM without exposing LibMR internal commands.
- Add a truth-table regression test for OSS, classic Enterprise, unmanaged native cluster, and ASM.

## Root cause

#98 treated every `cluster-enabled=yes` runtime as OSS. ASM also runs Redis with native clustering enabled, but loads ClusterPlugin with a configured topology URI. LibMR therefore selected its OSS HELLO/internal-secret transport instead of the Enterprise proxy-filtered transport. Peers rejected the internal HELLO command and fan-out operations timed out.

Both MOD-17016 support packages use LibMR `a37b672`, which contains #98 but predates #105. Their logs show the communication failure before scale begins, so this issue is independent of the topology repointing change in #105.

## Why this marker

In ASM, ClusterPlugin loads before data modules and registers `clusterplugin.cluster-topology-uri` as immutable. ClusterPlugin refuses to initialize while the value remains its default, `unset`. A resolved, non-default value therefore narrowly distinguishes an Enterprise-managed native cluster from the unmanaged private-binary OSS-cluster test setup.

The URI itself is never logged.

## Testing

- `make test_runtime_detection`
- `make run_tests` with Redis 7.4.10: standalone and 1-, 2-, and 3-shard OSS-cluster environments; zero failures in all four environments
- Private Enterprise Redis integration validation:
  - `cluster-enabled=yes` without ClusterPlugin selects OSS transport
  - `cluster-enabled=yes` with configured ClusterPlugin selects Enterprise transport and loads successfully with `_proxy-filtered` commands

Related: https://redislabs.atlassian.net/browse/MOD-17016

Context: https://github.com/RedisGears/LibMR/pull/98, https://github.com/RedisGears/LibMR/pull/105

---

> [!NOTE]
> **Medium Risk**
> Changes cluster init transport selection for multi-shard fan-out; mis-detection would still cause HELLO/timeouts or wrong command visibility, though behavior is covered by a truth-table test and preserves the #98 unmanaged-cluster case.
> 
> **Overview**
> **Fixes LibMR choosing the wrong inter-shard transport** when Redis Enterprise runs with native clustering (ASM): it now treats that layout as Enterprise and uses `_proxy-filtered` commands instead of the OSS HELLO/internal-secret path that peers reject.
> 
> Runtime classification is centralized in `MR_IsEnterpriseRuntime()` (`cluster_runtime.h`). Enterprise mode applies when `rlec_version` is present and either clustering is off **or** the shard is an **Enterprise-managed native cluster**. The latter is inferred when `cluster-enabled` is `yes` and `clusterplugin.cluster-topology-uri` is set to something other than the default `unset` (ASM loads ClusterPlugin with a real topology URI). The prior #98 behavior is preserved for an Enterprise binary in an **unmanaged** OSS-style cluster test setup (`cluster-enabled=yes` without that URI).
> 
> `MR_ClusterInit` refactors CONFIG reads through a shared `ConfigValueEquals()` helper (RESP2 array and RESP3 map), logs `managed-native-cluster`, and wires the decision into `clusterCtx.isOss` and command registration flags.
> 
> A small C truth-table test (`make test_runtime_detection`) is added and hooked into `run_tests`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64fb1f336357c9c9de0ee9bcd2dadcc8508fe15f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -12,6 +12,7 @@
 #include "common.h"
 #include "mr.h"
 #include "cluster.h"
+#include "cluster_runtime.h"
 #include "event_loop.h"
 #include "utils/arr_rm_alloc.h"
 #include "utils/buffer.h"
@@ -1757,6 +1758,42 @@ static void MR_NetworkTest(RedisModuleCtx *ctx, const char *sender_id, uint8_t t
     RedisModule_Log(ctx, "notice", "got a nextwork test msg");
 }
 
+static bool ConfigValueEquals(RedisModuleCtx *ctx, const char *name,
+                              const char *expected, bool *resolved) {
+    *resolved = false;
+    RedisModuleCallReply *reply = RedisModule_Call(ctx, "config", "cc", "get", name);
+    if (!reply) {
+        return false;
+    }
+
+    /* CONFIG GET replies as a flat array in RESP2 and as a map in RESP3. */
+    int replyType = RedisModule_CallReplyType(reply);
+    RedisModuleCallReply *value = NULL;
+    if (replyType == REDISMODULE_REPLY_ARRAY &&
+        RedisModule_CallReplyLength(reply) >= 2) {
+        value = RedisModule_CallReplyArrayElement(reply, 1);
+    } else if (replyType == REDISMODULE_REPLY_MAP &&
+               RedisModule_CallReplyMapElement &&
+               RedisModule_CallReplyLength(reply) >= 1) {
+        RedisModule_CallReplyMapElement(reply, 0, NULL, &value);
+    }
+
+    bool matches = false;
+    if (value && RedisModule_CallReplyType(value) == REDISMODULE_REPLY_STRING) {
+        size_t valueLen = 0;
+        const char *valueStr = RedisModule_CallReplyStringPtr(value, &valueLen);
+        if (valueStr) {
+            size_t expectedLen = strlen(expected);
+            *resolved = true;
+            matches = valueLen == expectedLen &&
+                      memcmp(valueStr, expected, expectedLen) == 0;
+        }
+    }
+
+    RedisModule_FreeCallReply(reply);
+    return matches;
+}
+
 int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
     clusterCtx.CurrCluster = NULL;
     clusterCtx.callbacks = array_new(MR_ClusterMessageReceiver, 10);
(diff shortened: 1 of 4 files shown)


src/cluster_runtime.h
```diff
@@ -0,0 +1,22 @@
...
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+ */
+
+#ifndef LIBMR_CLUSTER_RUNTIME_H
+#define LIBMR_CLUSTER_RUNTIME_H
+
+#include <stdbool.h>
+
+static inline bool MR_IsEnterpriseRuntime(bool rlecVersionPresent,
```
AvivDavid23, line 15: This is too complicated in my opinion. If we want a function that checks it, let it have the full check logic. Also we use it just once in the code, so more evidence that we should move all the logic in `src/cluster.c:1846` into here

src/cluster.c
```diff
@@ -1819,17 +1829,30 @@ int MR_ClusterInit(RedisModuleCtx* rctx, char *password) {
...
-     * no _proxy-filtered command flags) rather than the enterprise path. */
-    if (MR_RlecVersionPresent && !ossClusterRuntime) {
+     * are not in native cluster mode, except for an Enterprise-managed native
+     * cluster. In ASM, ClusterPlugin is loaded before data modules and registers
+     * an immutable clusterplugin.cluster-topology-uri config. ClusterPlugin
+     * refuses to load while the URI is "unset", so a non-default value narrowly
+     * distinguishes ASM from an enterprise binary running as a standalone OSS
+     * cluster (e.g. the env0 testing setup). */
+    bool managedNativeCluster = false;
+    if (MR_RlecVersionPresent && clusterEnabled) {
```
AvivDavid23, line 1839: check comment in MR_IsEnterpriseRuntime

tests/cluster_runtime_test.c
```diff
@@ -0,0 +1,29 @@
...
+ * Licensed under your choice of (a) the Redis Source Available License 2.0
+ * (RSALv2); or (b) the Server Side Public License v1 (SSPLv1); or (c) the
+ * GNU Affero General Public License v3 (AGPLv3).
+ */
+
+#include "../src/cluster_runtime.h"
+
+#include <assert.h>
+
+int main(void) {
```
AvivDavid23, line 14: I'm confused about out tests structure, not sure why the test module is in Rust... But anyway if we lack utest at all, I would just create a unit tests folder in `test` and put stuff there

AvivDavid23 (changes requested):
Check comments and resolve conflicts

